PR #74, opened 2014-12-30
Add C-Up/Down/Left/Right keys

amjith:
This change adds 4 more new key codes. C-Left, C-Right, C-Up and C-Down. 

It also adds C-Left and C-Right for Prev word and Next word in emacs bindings.

I've tested this change both with and without Tmux (which exercises the Application mode in tmux and Cursor mode outside of tmux). 

This change is currently targeted for VT100 only, since I don't have a windows machine to try and find the equivalent keycode in win32 console.


jonathanslenders:
Thank! I've merged it. 

I created a new issue for the Windows support (I haven't had time yet.):
https://github.com/jonathanslenders/python-prompt-toolkit/issues/78

PR #79, opened 2015-01-03
Fix jedi dependency to work withing the new PEP-440 rules

### LINKED ISSUE #72: DistributionNotFound exception with jedi version 0.8.1-final0
The most recent jedi version seems to be [`0.8.1-final0`](https://github.com/davidhalter/jedi/blob/cd648e933bfa89adcb2a92d856abe0b1900c0d90/jedi/__init__.py#L37), causing a `DistributionNotFound` exception during installation with setup.py or launching ptpython.
A possible workaround is to replace `jedi-0.8.1` with `jedi` in setup.py.

```
~$ pip3 freeze | grep -iP 'jedi|prompt'
jedi==0.8.1-final0
prompt-toolkit==0.25
```

ionelmc:
PEP-440 is implemented in pip 6.0.

Fixes #72.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -14,7 +14,7 @@ setup(
         packages=find_packages('.'),
         install_requires = [
             'docopt',
-            'jedi>=0.8.1',
+            'jedi>=0.8.1-final0',
             'pygments',
             'six>=1.8.0',
             'wcwidth',


jonathanslenders:
Thanks!

Would this fix this issue? https://github.com/jonathanslenders/python-prompt-toolkit/issues/72

ionelmc:
Yes, it would. Because `0.8.1-final0` is not valid version spec under the new pep440 rules it would trigger the legacy string comparing behaviour.

jonathanslenders:
Merged in https://github.com/jonathanslenders/python-prompt-toolkit/commit/8c2434adaacf8957f10220fbdcca529187f71dff

Thank you!

PR #88, opened 2015-01-17
Align lines of eval output to the 1st line in repl

pgy:
Hi,
Currently the lines in the `Out` blocks are not aligned, causing of an object with multi-line `repr` appear like this:
```python
In [1]: import numpy

In [2]: numpy.arange(16).reshape(4,4)
Out[2]: array([[ 0,  1,  2,  3],
       [ 4,  5,  6,  7],
       [ 8,  9, 10, 11],
       [12, 13, 14, 15]])

```

This PR makes the lines in the `Out` blocks aligned to the first one.

```python
In [1]: class MultilineRepr:
     2.     def __repr__(self):
     3.         return "three\nline\nrepr"  

In [2]: MultilineRepr()
Out[2]: three
        line
        repr
```

### DIFF
--- a/prompt_toolkit/contrib/repl.py
+++ b/prompt_toolkit/contrib/repl.py
@@ -113,16 +113,23 @@ class PythonRepl(PythonCommandLineInterface):
                 locals = self.get_locals()
                 locals['_'] = locals['_%i' % self.current_statement_index] = result
 
+                out_mark = 'Out[%i]: ' % self.current_statement_index
+
                 if result is not None:
                     try:
-                        self.stdout.write('Out[%i]: %r\n' % (self.current_statement_index, result))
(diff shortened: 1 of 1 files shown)


jonathanslenders:
Thanks for the pull request! I appreciate it.
I'll merge it soon. (A bit busy right now.)

jonathanslenders:
Merged in: https://github.com/jonathanslenders/python-prompt-toolkit/commit/28711f90c29e41322a36553da0d9a06291333fd1

(nice patch!)

pgy:
Thank you, but I've just spotted a bug in it :disappointed: . (`'%r\n' % result` should be `'%r\n' % (result, )` otherwise it'll break with tuples.) Later today or tomorrow I'll fix it.

PR #91, opened 2015-01-31
Fix example 3 code of the tutorial

jespino:


jonathanslenders:
Hi @jespino . Thanks for the fix. I really appreciate it. But I just realised that we would have to update this API change (we were using an old API there) in every example of this tutorial as well as in the sqlite-cli.py file in the same directory.

But actually, we are still using a low level API in there, it would become something like this now:
```python
layout = FloatContainer(                                                                                                                                                                                                                                                    
                Window(BufferControl(input_processors=[DefaultPrompt('> ')], lexer=SqlLexer),                                                                                                                                                                                   
                       menus=[CompletionsMenu()]),                                                                                                                                                                                                                              
                [                                                                                                                                                                                                                                                               
                    Float(xcursor=True, ycursor=True, content=CompletionsMenu(max_height=8))                                                                                                                                                                                    
                ]                                                                                                                                                                                                                                                               
            )   
```

Too complex for a tutorial that doesn't need to low level flexibility. I would rather make it this:

```python
from prompt_toolkit.contrib.shortcuts import get_input

while True:
    text = get_input('> ', completer=SqlCompleter(), style=DocumentStyle, lexer=SqlLexer)
    ....
```

jespino:
If you want, i can change the tutorial for use the shortcuts functions, can be good learning exercise :smile:

jonathanslenders:
If you have some spare time available, that would be very nice!

jonathanslenders:
I also see that you don't need to create a custom completer. The ``WordCompleter`` does everything we need in that example. The end result will probably look something like this:
```python
def main(database):                                                                                                                                                                                                                                                             
    connection = sqlite3.connect(database)                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                
    sql_completer = WordCompleter(['create', 'select', 'insert', 'drop',                                                                                                                                                                                                        
                                   'delete', 'from', 'where', 'table'], ignore_case=True)                                                                                                                                                                                       
    try:                                                                                                                                                                                                                                                                        
        while True:                                                                                                                                                                                                                                                             
            text = get_input('> ', completer=sql_completer,                                                                                                                                                                                                                     
                             style=DocumentStyle, lexer=SqlLexer)                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                                
            with connection:                                                                                                                                                                                                                                                    
                try:                                                                                                                                                                                                                                                            
                    messages = connection.execute(text)                                                                                                                                                                                                                         
                except Exception as e:                                                                                                                                                                                                                                          
                    print(repr(e))                                                                                                                                                                                                                                              
                else:                                                                                                                                                                                                                                                           
                    for message in messages:                                                                                                                                                                                                                                    
                        print(message)                                                                                                                                                                                                                                          
    except Exit:                                                                                                                                                                                                                                                                
        print('GoodBye!') 
```

And by testing this code, I just found a bug. ```get_input`` does not keep its history between successive calls. That's because we create a new ``History`` instance each time. I'll fix this last issue.

jespino:
Now better?

jespino:
Take care with the prompt_toolkit/contrib/shortcuts.py modification, this change is interesting for me, and is what the comment say :smile:

prompt_toolkit/contrib/shortcuts.py
```diff
@@ -190,7 +190,7 @@ def get_input(message='',
 
     # Read input and return it.
     on_abort = AbortAction.RAISE_EXCEPTION if raise_exception_on_abort else AbortAction.RETURN_NONE
-    document = cli.read_input(on_abort=on_abort, on_exit=AbortAction.IGNORE)
+    document = cli.read_input(on_abort=on_abort, on_exit=AbortAction.RETURN_NONE)
```
jonathanslenders: I will have a second look at this tomorrow. I'm not sure what the best behaviour is, actually.
If we look at what ``raw_input`` does for ``Ctrl-D`` and ``Ctrl-C``. it either throws ``KeyboardInterrupt`` or ``EOFError``.
Maybe not the most useful defaults, but somehow, I think if no parameters are given to ``get_input``, that it would be good if it behaves exactly as ``raw_input``.

What do you think?

(By the way: thank for the of the pull request! It looks good.)

jespino: I think is interesting to have some compatibility with the raw_input command to easy replace readline with prompt_toolkit, but you wont have this raising a different exception. If you raise KeyboardInterrput and EOFError anyone can change raw_input with get_input and the rest of the code must works.

I like the raise_exception_on_abort param, but with your idea of compatibility you have to change the default value to True, and (in my opinion) include another parameter for raise_exception_on_exit with default value to True. When the value is False, i think the most interesting option is RETURN_NONE.

jonathanslenders:
Hi @jespino I changed the ``get_input`` function to behave exactly as ``raw_input`` by default. Now it accepts an ``on_abort`` and ``on_exit`` attribute. I updated your pull request and merged it: https://github.com/jonathanslenders/python-prompt-toolkit/commit/6aee8ad086cfbfbfa883453b738512254b227783

Probably the tutorial can still be improved. (Not sure if everything is completely accurate, but I'll have a second look later on.)

PR #100, opened 2015-03-02
(Work in progress) Flush vt100 input after timeout.

jonathanslenders:
This should fix some issues with "playitagainsam" and Cygwin.

It's work in progress.


jonathanslenders:
Has been fixed already by another commit.

PR #107, opened 2015-04-28
remove comma

cs09g:


jonathanslenders:
I'm sorry. This is not an issue. Having the comma there has actually an advantage: you can sort the lines using the sort function of an editor without having to care about a missing comma in the middle.
It could be however that I'm not always consistent at having comma's at the end.

Anyway, I appreciate your pull request!

cs09g:
Cool, okay. Thanks.

PR #111, opened 2015-05-11
Fixing bug with default message in get_input()

anula:
Following code stopped working after last commit introducing `get_prompt_tokens` in `create_default_layout`:
```
from prompt_toolkit.shortcuts import get_input

if __name__ == '__main__':
    answer = get_input()
    print('You said: %s' % answer)
```
Because when message is empty (`''`) then code assumes that there is `get_prompt_tokens` function, which is not always true.

This is a very simple fix for this issue.

### DIFF
--- a/prompt_toolkit/shortcuts.py
+++ b/prompt_toolkit/shortcuts.py
@@ -115,7 +115,7 @@ def create_default_layout(message='', lexer=None, is_password=False,
     if extra_input_processors:
         input_processors.extend(extra_input_processors)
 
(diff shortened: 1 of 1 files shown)


jonathanslenders:
Thank you @anula!
This was my fault, thanks for fixing it!
It merged your pull request: https://github.com/jonathanslenders/python-prompt-toolkit/commit/4dfa286194a503339709e5e33d557bf82f410e9c

Cheers,
Jonathan

PR #112, opened 2015-05-13
Fix AssertionError in the layout example

sshin718:
When I try to run the layout example(layout.py), I get an error below.

```
  File "layout.py", line 144, in <module>
    main()
  File "layout.py", line 91, in main
    DefaultPrompt('python> '),
  File "/usr/lib/python3.4/site-packages/prompt_toolkit/layout/prompt.py", line 34, in __init__
    assert callable(get_tokens)
AssertionError
```

If it wants to gives only a message on initialization of `DefaultPrompt`, it should use `DefaultPrompt.from_message` instead.

### DIFF
--- a/examples/layout.py
+++ b/examples/layout.py
@@ -88,7 +88,7 @@ def main():
             Window(content=BufferControl(lexer=PythonLexer,
                                          show_line_numbers=Always(),
                                          input_processors=[
-                                                DefaultPrompt('python> '),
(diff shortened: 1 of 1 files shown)


jonathanslenders:
Thank you @sshin718 , I appreciate it.

Merged here: https://github.com/jonathanslenders/python-prompt-toolkit/commit/0deba118edcba0fbba8cd28bcdba374a02ebd911

PR #113, opened 2015-05-14
Add Ctrl-A & Ctrl-E support in vim insert mode.

obmarg:
These are shortcuts for home & end when in insert mode in vim. If your
keyboard doesn't have actual home and end keys, these are the only way
(that I know) to get to start/end of line without exiting insert mode.

### DIFF
--- a/prompt_toolkit/key_binding/bindings/vi.py
+++ b/prompt_toolkit/key_binding/bindings/vi.py
@@ -574,6 +574,22 @@ def load_vi_bindings(registry, vi_state, filter=None):
         # TODO: go to end of sentence.
         pass
 
+    @handle(Keys.ControlA, filter=insert_mode)
+    def _(event):
+        """
+        Move to first non whitespace of current line
+        """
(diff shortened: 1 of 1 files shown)


amjith:
Can we also add this to the visual selection mode? So I can navigate to home and end of a line while I'm selecting text using "Ctrl-V".

jonathanslenders:
I'm not sure about this actually...

In Vim, Control-E scrolls up and Control-A does nothing as far as I'm aware.

In readline. Vi insert mode. Control-E and Control-A are both inserted as literals ^E and ^A. In navigation mode, they seem to be ignored.

I agree that it's nice to have, but I prefer to not add too many customization to the key bindings.

Jonathan

obmarg:
Oh yeah - sorry, I've had that keybinding in my vimrc for so long that I forgot I added it myself.  Would not have submitted this PR if I'd realised.  Guess I can just add it using a custom config.py.  Thanks.

PR #114, opened 2015-05-20
Add 'M' keybinding to vi bindings.

amjith:
This adds the 'M' keybinding, which moves the cursor to the vertical center of the rendered lines in a buffer. 

It also adds 'z-' and 'z+' bindings. 

The end goal is to have this keybinding available in pyvim.

### DIFF
--- a/prompt_toolkit/key_binding/bindings/vi.py
+++ b/prompt_toolkit/key_binding/bindings/vi.py
@@ -954,6 +954,7 @@ def load_vi_bindings(registry, vi_state, filter=None):
         return CursorRegion(pos)
 
     @handle('z', '+', filter=navigation_mode|selection_mode)
+    @handle('z', Keys.ControlJ, filter=navigation_mode|selection_mode)
     def _(event):
         """
         Scrolls the window to makes the current line the first line in the visible region.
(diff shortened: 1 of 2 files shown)


jonathanslenders:
Thanks! This is great.

I really appreciate that.

jonathanslenders:
Thank you, all three commits have been merged. Works great!

anntzer:
Is this supposed to work e.g. in IPython 5?  I often end up wth the cursor toward the bottom of my terminal emulator (KDE/Konsole)'s scroll buffer, which makes tab completion slightly painful as it only shows whatever fits in the remaining ~5 lines or so.  A quick try suggests that none of z+, z- and M work there.  Thanks in advance.

jonathanslenders:
Hi @anntzer,

No, this is different. The key bindings in this commit will center the text (or move it up/down) in the current input region. If I recall correctly, they only make sense when the input contains already more lines then are available in the terminal height. So, this mostly makes sense for Pyvim.

However, what you are looking for definitely makes sense and it is possible to implement that. (Can you create a new issue for that?)

What key bindings would you like for this? It does not exist in GNU readline, so I'm not sure whether I will install default key bindings for this, but maybe we could do this anyway.

PR #122, opened 2015-06-02
Revise the typical usage of Renderer

dblN:
`sys.stdout` is not an instance of `Output`.


jonathanslenders:
Good catch! Thank you.

Merged in: https://github.com/jonathanslenders/python-prompt-toolkit/commit/d4b9deed6b43e55b64dbf5c45ded2e5cb70ebe0e

Issue #129, opened 2015-06-16
Control-T Emacs keybinding swaps wrong characters


brenns10:
In Emacs, the shortcut `C-t` (`transpose-chars`) does the following things:

- Swap the character under point with the character before it.
- Move point one character to the right.

(The only exception here is when point is at the end of the line, in which case it swaps the *two characters to the left of point*, and keeps point where it is).

So, if you had "a**b**c" in a buffer (bold represents point), then after doing `C-t`, you would have "ba**c**".

The keybinding in `prompt_toolkit` always swaps the two characters to the left of point, and doesn't move it.  It seems silly, but the transpose shortcut can be a ridiculously useful one, and when the semantics are just a bit off like this, it can really mess you up!  It would be nice if `C-t` behaved the same as Emacs.  Is this something that you'd feel comfortable changing, or is it implemented that way for a reason?

I see that the ControlT binding is implemented in `prompt_toolkit.key_bindings.bindings.basic` and not overridden in `prompt_toolkit.key_bindings.bindings.emacs`.  I could probably implement the expected behavior and submit a PR, but I wanted to figure out if you'd be open to such a change before spending time on it.

(Also, *great* work on this module!  I use it daily as a replacement for the default REPLs, and it blows them away!  I really appreciate the work.)

jonathanslenders:
Hi @brenns10,

Thanks a lot for pointing this out!
I wasn't aware of it (that's a shortcut a barely use myself) but it's clearly a bug. 

If you like to, feel free to submit a pull request and I'll merge it. (Please fix in ``key_bindings.bindings.basic``.)

Cheers and thanks for the complement,
Jonathan

PR #130, opened 2015-06-17
Correct emulation of Emacs' C-t (transpose-char) behavior.

brenns10:
This addresses my concern in issue #129.  I tested this with single and multiline input using the examples, and everything behaves as expected.  However, if there's a standard way to test these things, I can add unit tests too.

### DIFF
--- a/prompt_toolkit/key_binding/bindings/basic.py
+++ b/prompt_toolkit/key_binding/bindings/basic.py
@@ -197,7 +197,7 @@ def load_basic_bindings(registry, filter=Always()):
             return
         elif p == len(b.text) or b.text[p] == '\n':
             b.swap_characters_before_cursor()
-        elif b.document.get_start_of_line_position:
+        else:
             b.cursor_position += b.document.get_cursor_right_position()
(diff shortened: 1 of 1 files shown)


jonathanslenders:
Hi @brenns10,

Thanks you very much! It's merged in: https://github.com/jonathanslenders/python-prompt-toolkit/commit/c20a09f7f497b5a53e71d9ff4ae6a6e79cf2140e

Also, thanks for offering to write tests. However, unfortunately none of the other key bindings are currently tested, and I'm really not sure what the best way to test it is. (For Vi mode for instance, keys are composed. You have for instance 'w', 'dw', 'yw', 'cw', etc.. but these combinations are generated. So testing each individual combination and edge case would be a whole lot more code then the code for the key bindings itself.) Maybe it's easy to have 100% coverage, but in reality, that covers only a small fraction of the possible flows and bindings. So I propose to leave it for now.

I guess in the future we will maybe test the most common ones and keep the focus rather on the data structures on which these key bindings operate. In the end, there should be not too much code in the key binding handlers.

Jonathan

PR #151, opened 2015-08-28
Fix tutorial using abstract History type (fixes #150)

### LINKED ISSUE #150: TypeError when running tutorial
When running `sqlite-cli.py`, it fails with

      File "sqlite-cli.py", line 55, in <module>
        main(db)
      File "sqlite-cli.py", line 29, in main
        history = History()
    TypeError: Can't instantiate abstract class History with abstract methods __getitem__, __len__, append

I'm running Python 3.4.3

ALSchwalm:
I went ahead and just removed the explicit usage of History. I'm not sure if you'd prefer to use FileHistory to show that off. If so, I can change it to use that.

### DIFF
--- a/examples/tutorial/README.rst
+++ b/examples/tutorial/README.rst
@@ -37,16 +37,16 @@ Let's get started!
 #. Loop The REPL
 
    Now we call the ``get_input`` method in a loop while keeeping the line
-   history in our ``History`` object.
+   history in our ``InMemoryHistory`` object.
 
    .. code:: python
 
        from __future__ import unicode_literals
        from prompt_toolkit.shortcuts import get_input
-       from prompt_toolkit.history import History
+       from prompt_toolkit.history import InMemoryHistory
 
        def main():
-           history = History()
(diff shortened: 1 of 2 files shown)


jonathanslenders:
Hi @ALSchwalm, thanks for the pull request!
I guess it's better to create an ``InMemoryHistory`` instance instead. The reason is that otherwise, for each call of ``get_input``, a new ``InMemoryHistory`` instance is created, and basically that way, there is no history between consecutive calls of ``get_input``.

ALSchwalm:
Yep, I should have tested this a bit more thoroughly. Fixed and rebased. Thanks for your work on this project, by the way. I'm working on some stuff with it and It's been a pleasure to use.

jonathanslenders:
Thanks you! Merged in: https://github.com/jonathanslenders/python-prompt-toolkit/commit/c1170c9b72eec9f1b57fd96a515dc2609bacfda6

(I also appreciate the rebase ;) )

PR #152, opened 2015-09-04
MANIFEST.in to include LICENSE and CHANGELOG

lhw:
MANIFEST.in to include LICENSE and CHANGELOG in sdist for pypi

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,0 +1,3 @@
+include *rst LICENSE CHANGELOG
+recursive-include examples *.py
+prune examples/sample?/build


kitterma:
It would be nice to have MANIFEST.in also include MANIFEST.in so the sdist can be rebuilt from the sdist (this is good for downstream distributors like Linux distributions).

jonathanslenders:
Hey @kitterma,

Thanks for the pull request. Is there any place where I can find more information about this manifest file, and why it's important to have one?

Thanks,
Jonathan

kitterma:
I didn't do the original pull request (lhw did), but I can speak to what he was trying to accomplish as we are collaborating to get prompt_toolkit packaged in Debian.
There are two files (LICENSE and CHANGELOG) that it would be good to have in the sdist, each for their own reason:
We want LICENSE because since Debian distributes both source and binary we want a copy of the exact license for the code in our source distribution so the the requirements are clear and self-contained.  I think this is a good general practice anyway.
We want CHANGELOG so we can ship it in the package documentation to enable users to see what has changed over time with the package.

MANIFEST.in is just a way to add files to the sdist (it's the normal way in distutils).  I'm not that versed in setuptools myself, but I do know there are other ways to do it.  What's important (at least from our point of view) isn't the MANIFEST.in file itself, but the added files it would add to the sdist.

If the MANIFEST.in isn't shipped with the sdist, then a downstream distributor that modified the package might get a different result.  I believe it's a good general practice to include all the components of a package build system when you ship it.

That's probably way more information than you wanted ...

Scott K

jonathanslenders:
Thanks for the explanation. I appreciate it!

I merged the pull request from @lhw in this commit: https://github.com/jonathanslenders/python-prompt-toolkit/commit/0fbe8ee86d33699f87b43706302ac6a56aeb3ddb

Can you create a new pull request with the required changes?

PR #153, opened 2015-09-04
Update MANIFEST.in

kitterma:
Include MANIFEST.in in the files shipped in the sdist so the sdist can be rebuilt from itself.


jonathanslenders:
Thank you. Merged in: https://github.com/jonathanslenders/python-prompt-toolkit/commit/21c0b609ae6dfae2d4d39158a0e3e378392218b5

PR #155, opened 2015-09-06
Auto suggestion

jonathanslenders:
Working implementation of auto-suggest for prompt-toolkit shells.

@amjith: Have a look if you are interested. It's working and finished, but I'm going to give this a few more days in my head to see whether something can be even better.

### DIFF
--- a/examples/auto-suggestion.py
+++ b/examples/auto-suggestion.py
@@ -1,0 +1,40 @@
+#!/usr/bin/env python
+"""
+Simple example of a CLI that demonstrates fish-style auto suggestion.
+
+When you type some input, it will match the input against the hitory. If One
+entry of the history starts with the given input, then it will show the
+remaining part as a suggestion. Pressing the right arrow will insert this
+suggestion.
+"""
+from __future__ import unicode_literals, print_function
+from prompt_toolkit.shortcuts import get_input
+from prompt_toolkit.history import InMemoryHistory
+from prompt_toolkit.interface import AbortAction
+from prompt_toolkit.auto_suggest import AutoSuggestFromHistory
+
+
+def main():
+    # Create some history first. (Easy for testing.)
+    history = InMemoryHistory()
+    history.append('import os')
+    history.append('print("hello")')
+    history.append('print("world")')
+    history.append('import path')
+
+    # Print help.
+    print('This CLI has fish-style auto-suggestion enable.')
+    print('Type for instance "pri", then you\'ll see a suggestion.')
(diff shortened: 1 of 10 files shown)


amjith:
This works as I'd expect. A few minor nit-picks: 

1. C-f should also fill in the command since it is an alternate key binding for right arrow. 
2. `pri` and up-arrow should work like search where it starts to cycle through the history entries that have `pri` in them. 
3. I'd make the suggestion color a little more lighter. Right now it is more prominent than the actual entered text. 
eg:

1__1_1_python2_7_-__python___users_amjith_dropbox_code_python_python-prompt-toolkit___tmux_

jonathanslenders:
Hi @amjith,

1. The Ctrl-F key binding has been added.
2. The example has been updated that it enables this behaviour. It can be enabled separately. (`` enable_history_search ``).

3. This is difficult. Probably you changed your terminal to use light gray as the foreground color and reduce the contrast a litte. The terminal however only changes the color for the "default" color which is normally either black or white, depending on the settings. For the history suggestion, I use ``#448888`` which maps to a certain gray with a little bit of blue on the 356 color pallette. This is fixed, and doesn't adapt to the terminal colors. I tried to choose something that is darker than white on a white-on-black terminal, but also lighter than black on a black-on-white terminal. So, basically gray. I added a tiny bit of blue, in order to make sure that there will always be some difference, even in case the foreground is a bit gray.

    There is one alternative. That's to use the ANSI "dim"  sequence (``\e[2m``). Unfortunately Pygments doesn't support it, and also many terminals and tmux doen't support it. So, that's not really an option.

    But basically, for the perfect color, we have to know whether the background is black or white and there is no way to detect that automatically.

One more question regarding naming. For autocompletion, a class that generates the completions is called a ``Completer``. What would be the equivalent for something that creates auto suggestions? I called the class right now ``AutoSuggest``.

Jonathan

amjith:
Ctrl-F works and the history search is also working. 

Regarding the colors: This is what it looks like in Fish. That gray seems to work good for both dark background and light background. 
1__elm-dev_2_fish_-__fish___users_amjith_dropbox_code_python_python-prompt-toolkit___tmux_

I took an eye dropper tool to that color and it looks like it uses RGB 102,102,102.

I think the class name AutoSuggest is great. I noticed that the history specific sub-class is called AutoSuggestFromHistory which is also a good name.

jonathanslenders:
Cool. I updated the default styling to use 102,102,102 and merged it.

PR #159, opened 2015-09-19
Use call_soon_threadsafe() in call_from_executor().

niklasf:
`call_from_executor()` is inteded to be called from a different thread. Thus use the threadsafe version of `call_soon()`: [`call_soon_threadsafe()`](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.BaseEventLoop.call_soon_threadsafe).

### DIFF
--- a/prompt_toolkit/eventloop/asyncio_posix.py
+++ b/prompt_toolkit/eventloop/asyncio_posix.py
@@ -100,4 +100,4 @@ class PosixAsyncioEventLoop(EventLoop):
         Call this function in the main event loop.
         Similar to Twisted's ``callFromThread``.
         """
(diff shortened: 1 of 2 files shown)


jonathanslenders:
Thank you @niklasf!
This was indeed not correct. I wonder actually how you discovered this bug and whether it really broke some stuff.

Merged in: https://github.com/jonathanslenders/python-prompt-toolkit/commit/2e28beec13901720bb103ae490c6cda6ff5c8d15

niklasf:
I was using an application that had bursts of logging output coming from a thread. It caused mixed output and broken terminal escape codes. (Of course only because I was using call_from_executor() to actually do output. See follow-up pull request.)

PR #160, opened 2015-09-19
Make the stdout proxy threadsafe

niklasf:
This is a bit more controversial than the last one. But I think it's generally a good idea to make the stdout proxy threadsafe.

* Calls to _StdoutProxy.write() and _StdoutProxy.flush()
  are now atomic.

* Since CommandLineInterface.run_in_terminal() is not
  threadsafe, use call_from_executor() to call it from the
  stdout proxy.

### DIFF
--- a/prompt_toolkit/interface.py
+++ b/prompt_toolkit/interface.py
@@ -3,10 +3,12 @@ The main `CommandLineInterface` class and logic.
 """
 from __future__ import unicode_literals
 
+import functools
 import os
 import signal
 import six
 import sys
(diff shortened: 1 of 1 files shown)


jonathanslenders:
Hi @niklasf. I agree on this. ``stdout`` is a global object, and every thead should be able to write to it. (That's what people expect.)

Merged in: https://github.com/jonathanslenders/python-prompt-toolkit/commit/bce2b85826bcc2053cee9a62356a6cc677ee3d7b

PR #165, opened 2015-09-25
ignore termios errors on get/set terminal attributes

ddalex:
Not all terminals where prompt_toolkit is used support get/set
attributes - a glaring example is pipe-ing in, e.g. executing
a pipe over SSH where no full terminal is allocated

This patch allows prompt_toolkit to continue working in such
environments by simply not attempting to change term attributes
where this is used.

Tested with pipeing through mycli.

### DIFF
--- a/prompt_toolkit/terminal/vt100_input.py
+++ b/prompt_toolkit/terminal/vt100_input.py
@@ -341,22 +341,29 @@ class raw_mode(object):
     """
     def __init__(self, fileno):
         self.fileno = fileno
-        self.attrs_before = termios.tcgetattr(fileno)
+        try:
+            self.attrs_before = termios.tcgetattr(fileno)
+        except termios.error:
+            self.attrs_before = None
 
     def __enter__(self):
         # NOTE: On os X systems, using pty.setraw() fails. Therefor we are using this:
-        newattr = termios.tcgetattr(self.fileno)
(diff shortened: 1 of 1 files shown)


jonathanslenders:
Hi @ddalex, thanks for pointing this out and taking the time to write a patch! I appreciate it.

Just to be clear that I understand correctly. This is to be used when stdin/stdout are not attached to a pty device, right? So in that case, we can send the output raw. Is it correct to assume that the terminal is already raw? E.g. If I run ``cat - | ptpython``, this patch would avoid the crash, but it doesn't make the terminal raw.

What's your opinion of using ``os.isatty`` for this? I would tend to check in ``prompt_toolkit.input.StdinInput.raw_mode/cooked_mode`` whether the given stdin object is a terminal, and if not, return a dummy context manager. (E.g. ``contextlib.nested()``.) I'm not sure whether it solves the same issue or has any possible disadvantages.

If I remember correctly, I believe there's also the concept of the "controlling terminal" of a process. I think it has something to do with the process group. But I guess that doesn't help anything.

jonathanslenders:
This was implemented slighty different in: https://github.com/jonathanslenders/python-prompt-toolkit/commit/865dafbfae4da5f7dae364d2e9310942e1abd77c
Thanks again for this pull request!

PR #176, opened 2015-10-11
Vi movement bindings 't' and 'f' were off by one.

amjith:
Ref: https://github.com/dbcli/mycli/issues/167

'dt' and 'df' are off by one in the Vi mode.

### DIFF
--- a/prompt_toolkit/key_binding/bindings/vi.py
+++ b/prompt_toolkit/key_binding/bindings/vi.py
@@ -839,7 +839,7 @@ def load_vi_bindings(registry, vi_state, enable_visual_key=Always(), filter=None
         """
         vi_state.last_character_find = CharacterFind(event.data, False)
         match = event.current_buffer.document.find(event.data, in_current_line=True, count=event.arg)
-        return CursorRegion(match or 0)
+        return CursorRegion((match + 1) if match else 0)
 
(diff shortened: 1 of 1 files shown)


jonathanslenders:
Thanks @amjith for pointing out the issue.

However, the patch is not completely correct. It fixes df and dF, but breaks f again.

It's supprising actually, because "dw" for instance does not delete the target character of "w", nor does it happen with any of the other key bindings.

 I think it would be better to give the ``CursorRegion`` class an additional parameter, ``include_character_below_cursor`` or something like that, which when True, also be copies/cut/delete the target character.

amjith:
You're right. The patch is not quite complete. 

I was trying to change the [`create_transform_handler`](https://github.com/jonathanslenders/python-prompt-toolkit/blob/master/prompt_toolkit/key_binding/bindings/vi.py#L620) function to handle this case but it had no effect.

jonathanslenders:
This has been fixed. (Not yet released on Pypi.)
Most of-by-one stuff has been fixed in: https://github.com/jonathanslenders/python-prompt-toolkit/commit/f56a75a96160b5ee3112ee1603eaf4087a09d092
Thanks to @yyt16384.

So now we also have separate key bindings for Vi text objects (motions) and operators.

Issue #182, opened 2015-11-04
Float xcursor/ycursor


Carreau:
I think I understand the usage of `(x|y)cursor` in Float that are after used [here](https://github.com/jonathanslenders/python-prompt-toolkit/blob/a864f6eff3a24f863e9b0cfae7ded42f5ba90c80/prompt_toolkit/layout/containers.py#L276-L381) in `write_to_screen `, do you think it could be a better/simpler idea to have a singleton `cursor_pos = object()` (or even a specific string `"cursor"`) that could could be use for `Float`s top/left/bottom/right arguments ?

The advantage of the `Cursor` singleton object is you could implement `__add__/__sub__` dunder, 

And do things like:

```
from prompt_toolkit.shortcut import cursor as c
Float(left=c+1, top=c-3)
```

Thoughts?

ljluestc:
from prompt_toolkit.layout import Float
from prompt_toolkit.formatted_text import HTML


class Cursor:
    instance = None

    def __new__(cls):
        if cls.instance is None:
            cls.instance = super(Cursor, cls).__new__(cls)
        return cls.instance

    def __add__(self, value):
        return self

    def __sub__(self, value):
        return self


cursor = Cursor()

def write_to_screen(text, cursor_pos, color):
    return [
        Float(
            top=cursor_pos,
            left=cursor_pos,
            content=HTML(text),
            background_color=color,
        )
    ]

layout = write_to_screen("Hello", cursor + 1, "red") + write_to_screen("World", cursor - 3, "blue")

PR #183, opened 2015-11-04
Use singleton object for Float x/ycursor

Carreau:
Attempt to start implementing #182

-- 
I'm not familiar with the codebase so it might not be he right way to do that, 
I suppose the all logic here could be simplified and keep backward compat with `(x|y)cursor`

Screenshot with this as a completer:
```
Float(left=cursor+7,
       top=cursor-1,
                      ...))
```

screen shot 2015-11-03 at 20 48 48

(note, my goal is to have a completer like Fish which is a `cursor.y+1` but the all width of terminal in the end, but not really a hsplit)

### DIFF
--- a/prompt_toolkit/layout/containers.py
+++ b/prompt_toolkit/layout/containers.py
@@ -24,7 +24,8 @@ __all__ = (
     'Window',
     'WindowRenderInfo',
     'ConditionalContainer',
-    'ScrollOffsets'
+    'ScrollOffsets',
+    'cursor',
 )
 
 Transparent = Token.Transparent
@@ -284,6 +285,8 @@ class FloatContainer(Container):
                                 y=cursor_position.y - write_position.ypos)
 
         for fl in self.floats:
+            #if isinstance(fl.left, Cursor):
+            #   fl.left = fl.left.delta+cursor_position.x
             # Left & width given.
             if fl.left is not None and fl.width is not None:
                 xpos = fl.left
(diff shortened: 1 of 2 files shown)


jonathanslenders:
Hi @Carreau,

Thanks a lot for the issue and the work on the pull request!

My thoughts:
- In general, I very much like the idea of the Cursor singleton and the ``__add__``/``__sub__`` overload.
- However, this is still not going to be flexible enough. The current algorithm will position the content of the float above the cursor when the input spans multiple lines. In that case, saying ``top=cursor-1`` is incorrect.

Possibly a better approach is to rewrite the ``FloatContainer.write_to_screen`` function and use a more powerful interface for plugging in other float positioning algorithms.
Ptpython for instance, also uses the ``BufferControl.menu_position`` to display the function signature at the start of the function call. (https://github.com/jonathanslenders/python-prompt-toolkit/blob/master/prompt_toolkit/layout/controls.py#L457)  I would like to get rid of that, if we find a better, more generic approach.

For this, I also don't mind to be backwards-incompatible (it's mostly used internally), but I prefer to do it correctly.

Currently I'm still abroad for work with few spare time, but I will think about this. It's definitely interesting!

Jonathan

Carreau:
> In that case, saying top=cursor-1 is incorrect.

Yeah, that was just an example to see that `minus` was working definitively not good for UI. 

> For this, I also don't mind to be backwards-incompatible (it's mostly used internally), but I prefer to do it correctly.

I'm still wrapping my head around the different case here. Especially xcusrsor and ycursor are hints for positioning. I'm trying to find the correct abstraction. To simplify the interface. 

I'm also pretty busy with IPython these days, I was mostly trying to change my mind a bit, but will try to work a bit on that and understand more of the code.

PR #198, opened 2015-12-14
Respect _StdoutProxy.raw when flushing

niklasf:
Hi, the new `_StdoutProxy.raw` is a really useful feature. It should also be respected when flushing the buffer.


jonathanslenders:
Thank you for solving this bug, @niklasf!
I appreate it.
Merged in: https://github.com/jonathanslenders/python-prompt-toolkit/commit/ef89d7d73cd9989f581d4c14c02b862d724c6b4f
Jonathan

PR #210, opened 2016-01-06
Implementation of named registers.

jonathanslenders:
Implementation of named registers.

This works, but it increases the start-up time significantly.
A better solution is recommended.


jonathanslenders:
Implemented in https://github.com/jonathanslenders/python-prompt-toolkit/commit/1ab9ad0e86b612be1e949eb8c965072eb1d566ad

PR #215, opened 2016-01-14
setup.py - add trove classifiers and version parsing

adamchainz:
Found that `prompt_toolkit` is not listed as Python 3 compatible by `caniusepython3`, because it's lacking Trove Classifiers. Added them. Due to adding a License Trove Classifier, it's no longer necessary to provide the `license` argument to `setup`.

Whilst I was at it I saw you have the version defined in two files with a comment to remind you about updating both. I have adapted some code I use on a number of projects to ease this burden so the version is only stored in `__init__`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,5 +1,6 @@
 #!/usr/bin/env python
 import os
+import re
 from setuptools import setup, find_packages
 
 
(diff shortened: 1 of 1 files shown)


jonathanslenders:
Thank you @adamchainz,
Merged in: https://github.com/jonathanslenders/python-prompt-toolkit/commit/8a0e11e6ef77a129dc795469aefa36f3ad10ce00

PR #216, opened 2016-01-14
Travis - build faster

adamchainz:
This activates Travis' container based infrastructure ( https://docs.travis-ci.com/user/workers/container-based-infrastructure/ ) which is significantly faster for smaller projects like this.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+sudo: false
 language: python
 env:
   - TOXENV=py26


jonathanslenders:
Thank you, @adamchainz.
I was not aware of this, but that's nice! Merged in: https://github.com/jonathanslenders/python-prompt-toolkit/commit/3b1d668182cbf5615ac7b557b4dc200be7b130fd

PR #219, opened 2016-01-15
Add softlayer python to list of prompt-toolkit users

kitterma:
This is the only user of prompt-toolkit that's packaged in Debian that wasn't already listed.


jonathanslenders:
Thank you! Merged in: https://github.com/jonathanslenders/python-prompt-toolkit/commit/7bd72d81b2ce5520e768606a6cd9e08ace24301c

Issue #206, opened 2016-01-02
Completion doesn't insert common prefix


evanunderscore:
When running `examples/autocompletion.py`, if I attempt to complete an animal starting with 't', the options are 'turkey' and 'turtle'. In this instance, I would expect 'tur' to be filled in before I start tabbing through the options.

It's unclear to me what the intended behavior of this is. There appears to be code to do this in interface.py (`insert_common_part`), and it looks like this variable is getting set to True, but that code block never seems to get hit.

For reference, I'm using Python 2.7 on Windows 10.

evanunderscore:
I just took another look at this. The problem seems to be that pushing `t` immediately brings up the completions menu, so the first tab starts cycling through it. If I push another letter then backspace, the completion box goes away, and pushing tab fills in the common prefix as expected (but again shows the completion box immediately, counter to the description in the module docstring).

PR #226, opened 2016-01-25
Fixed examples/autocompletion.py

evanunderscore:
Fixes the issue described in #206 (somewhat).

I've added `completion_while_typing=False` to mostly match the module docstring, then updated it to describe what the actual behavior currently is.

Things I'm still unsure about:
* Is there a way to get the behavior described in the docstring as it was?
* Is there an option to insert a space after a full match? Currently the cursor sticks to the end of the word and the menu shows just that one option.

### DIFF
--- a/examples/autocompletion.py
+++ b/examples/autocompletion.py
@@ -3,8 +3,8 @@
 Autocompletion example.
 
 Press [Tab] to complete the current word.
-- The first Tab press fills in the common part of all completions.
-- The second Tab press shows all the completions. (In the menu)
+- The first Tab press fills in the common part of all completions
+    and shows all the completions. (In the menu)
 - Any following tab press cycles through all the possible completions.
 """
 from __future__ import unicode_literals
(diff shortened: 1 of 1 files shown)


jonathanslenders:
Thank you! Merged in https://github.com/jonathanslenders/python-prompt-toolkit/commit/9612b3e7651cf2042047bee9819c62f8434ee8e0

For the first question. I'm not sure right now. (I will respond later.)

For the second question. Yes. When you create a ``Completion`` instance. Make sure that the ``text`` attribute has a space at the end, but the ``display`` attribute doesn't. I'm not sure that the ``WordCompleter`` class supports it correctly right now, but the completer API should.
https://github.com/jonathanslenders/python-prompt-toolkit/blob/master/prompt_toolkit/completion.py#L15

jonathanslenders:
I think I can close this. Feel free to reopen if not. (Or create a new ticket.)

evanunderscore:
Thanks. I'll play around with it and open a new ticket if I need to.

PR #227, opened 2016-01-25
Disable bracketed paste support in ConEmu

BYK:
ConEmu supports bracketed paste but prompt-toolkit does not process bracketed pastes on Windows properly so disable it.

This patch fixes the issue where you get `[200~` and `[201~` around pasted strings when using ConEmu.

### DIFF
--- a/prompt_toolkit/terminal/conemu_output.py
+++ b/prompt_toolkit/terminal/conemu_output.py
@@ -32,7 +32,8 @@ class ConEmuOutput(object):
     def __getattr__(self, name):
         if name in ('get_size', 'get_rows_below_cursor_position',
                     'enable_mouse_support', 'disable_mouse_support',
-                    'scroll_buffer_to_prompt', 'get_win32_screen_buffer_info'):
+                    'scroll_buffer_to_prompt', 'get_win32_screen_buffer_info',
+                    'enable_bracketed_paste', 'disable_bracketed_paste):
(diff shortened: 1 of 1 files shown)


jonathanslenders:
Thank you for this, @BYK!
I'll try to merge it soon.

(If bracketed paste is supported, is there another way to implement it?)

BYK:
@jonathanslenders - no problem!

I actually tried to implement bracketed paste in win32 but the architecture of it is quite different from vt100 so it hardly applies. I can argue `conemu` should use `vt100_input` but my quick attempt to replace that failed with a `termios module not found` error.

BYK:
@jonathanslenders - hey there, it's been a while. Anything I can make to move this PR forward?

jonathanslenders:
Hi @BYK,

Sorry for the delay, and thanks for reminding me. I merged it: https://github.com/jonathanslenders/python-prompt-toolkit/commit/ce9f80f00fe5a121dc292471be0c223879dcd4c0

Now, I also realise the reason. The output goes through the Vt100 class, but the input still goes through the win32 input class. I will have to think of a workaround in order to support real bracketed paste on ConEmu.

Jonathan

BYK:
Yay, thanks!

BYK:
Also let me know if I can help with supporting bracketed paste in ConEmu. I rely on your project, ConEmu and Xonsh quite heavily and would love to be able to help out.

PR #228, opened 2016-01-28
Improve the documentation for beginers

renanivo:
* Add `unicode_literals` import in "Getting Starded" code snippet
  
  I think it is important since I copied and pasted the code from the _"Getting Started"_ and it didn't work in python 2.7 (And I got just an `AssertionError`). I know there's a note about it in the _"Building prompts"_ section, but I already figured out the problem when I got there.

* Link to the stable version of the documentation

  There are some functions in the latest version of the docs that aren't available in the current version of the project (0.57).

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -7,7 +7,7 @@ Python Prompt Toolkit
 and terminal applications in Python.
 
 Read the `documentation on readthedocs
-<http://python-prompt-toolkit.readthedocs.org/en/latest/>`_.
+<http://python-prompt-toolkit.readthedocs.org/en/stable/>`_.
 
 
 Ptpython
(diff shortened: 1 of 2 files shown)


jonathanslenders:
Thank you @renanivo,

Merged in https://github.com/jonathanslenders/python-prompt-toolkit/commit/4a3aa7d6a7c41dd82f5b3a273d41b161aa321ec4 and https://github.com/jonathanslenders/python-prompt-toolkit/commit/f0656ea6d18bffb8560c1fe9658405196f33c63c

renanivo:
You're welcome. Congratulations for your project and thanks for your quick reply.

PR #233, opened 2016-02-07
Refactor layout window

jonathanslenders:
Refactoring of the layout engine.
This should allow a BufferControl to be at least 100x larger. (Editing a buffer with about 200,000 lines works fine now.)

TODO:
- scroll offsets when line wrapping is on. (+ maybe key bindings.)
- cleanup WindowRenderInfo
- Test wide character (Asian) input


jonathanslenders:
Merged in other commit.

PR #238, opened 2016-02-13
Big refactoring of the Window and user control interfaces.

jonathanslenders:
Huge performance improvements for big inputs.

The Window and UIControl have been rewritten very much. Rather than each time
rendering the whole user control, we now only have to render the visible part.
(Saving a lot of CPU.)

Because of this, many pieces had to be rewritten:
- UIControls work differently. They return a UIContent instance that consist
  of a collection of lines.
- All processors have been rewritten.
- Some bugs have been fixed.


jonathanslenders:
(Merged in other commit.)

PR #240, opened 2016-02-14
Allow ansi color name in Pygments style.

Carreau:
Hey, 

I know that currently you can use vt100 to get GREN/BLUE ... etc, though if you create your own style that should work with pygments, it wont be accepted as these string won't be detected as correct colors.

Pygments recently added the ability to do that, except the names are slightly different, so this (once finished) should allow to have themes that works correctly with both Pygments 2.2+  and Prompt Toolkit **and** get the correct ansi escape sequences.

PR #242, opened 2016-02-15
Update to match color rename to match pygments

kitterma:
I think this will fix the test failure.


Carreau:
I think it is slightly more complicated (see #240), and this will only work on pygments master for now.

jonathanslenders:
I guess we can now close this. Is that right, @Carreau?

Carreau:
Yes, I guess.

PR #243, opened 2016-02-16
Start rewriting the full screen application example.

Carreau:
Use an order which allow to experiment while reading instead of having
all fall in place at once in the end.

### DIFF
--- a/docs/pages/full_screen_apps.rst
+++ b/docs/pages/full_screen_apps.rst
@@ -8,8 +8,101 @@ applications. Typically, an application consists of a layout (to describe the
 graphical part) and a set of key bindings.
 
 The sections below describe the components required for full screen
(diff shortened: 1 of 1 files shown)


jonathanslenders:
Just to be sure. This is ready to be merged, right?

After the next prompt-toolkit release (hopefully somewhere next week), I'm planning to invest on extending the documentation more, because there's still a lot missing.

In any case, thanks for doing this. I appreciate it.

Carreau:
> Just to be sure. This is ready to be merged, right?

Yes, I'll try to do it by small chunks. Still trying to learn it. 
Will send more refactor later likely.

jonathanslenders:
Thanks again Matthias! Merged in https://github.com/jonathanslenders/python-prompt-toolkit/commit/704e694533c3b60de4eecfb207d7c15b842a5abe

PR #244, opened 2016-02-18
Changed dict check to Mapping

scopatz:
This is causing problems for some proposed xonsh changes.  Also there is no need to check against concrete types, usually. See https://github.com/scopatz/xonsh/pull/687#issuecomment-185802948

Note that we use ChainMap in our pygments Style class because of the need to keep different color namespaces separate.


jonathanslenders:
Thank you Anthony, this is merged in https://github.com/jonathanslenders/python-prompt-toolkit/commit/eab5b72ee29291c1e67dc62b14e8ca230ace552c. Definitely makes sense.

scopatz:
Thanks!

PR #246, opened 2016-02-21
Remove unnecessary comment for AcceptAction

38elements:
`render_ci_done` is not an argument of `AcceptAction()`.

```
    def __init__(self, handler=None):
        assert handler is None or callable(handler)
        self.handler = handler
```


jonathanslenders:
Thank you @38elements!
This is merged in: https://github.com/jonathanslenders/python-prompt-toolkit/commit/fab6c92c1ef7a6f79d3414b1f2e7988dbfe4b148

PR #253, opened 2016-02-26
Improve Vi mode.

yyt16384:
Previously all motions are treated as exclusive motions.

Also fix incorrect behavior of some motions.

I found out that `Document.find_start_of_previous_word` and `Document.find_previous_word_beginning` are the same. Should I remove one of them?

### DIFF
--- a/prompt_toolkit/document.py
+++ b/prompt_toolkit/document.py
@@ -502,7 +502,7 @@ class Document(object):
     def find_previous_word_beginning(self, count=1, WORD=False):
         """
         Return an index relative to the cursor position pointing to the start
-        of the next word. Return `None` if nothing was found.
+        of the previous word. Return `None` if nothing was found.
         """
         regex = _FIND_BIG_WORD_RE if WORD else _FIND_WORD_RE
         iterator = regex.finditer(self.text_before_cursor[::-1])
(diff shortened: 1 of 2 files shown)


jonathanslenders:
Thanks for this @yyt16384! This is really great.
(I'll have to wait at least until Sunday before I can review the code. But then - if everything is good - I'll merge it.)

jonathanslenders:
Just a message to tell that I didn't forget about this. Just very busy. But still planning to review and merge it.

yyt16384:
I just updated this to fix some small problems.

Also, I changed the title to "Improve Vi mode" because it also fixes some other problems with Vi mode. They are listed in the commit message. The Vi mode is more broken than I thought...

There are still some problems that I can't fix without making big changes. Like not being to use a count between operator and motion (`d2j` won't work). Ideally we should use a operator pending mode, which will also reduce the number of keybindings and allow users to add new operators and motions more easily.

jonathanslenders:
Cool! Let me have a look at this first, and then we'll see how to go on from there.

I'm definitely in favour of adding more modes. Both to make it easier to add new operators, as well for keeping the amount of key bindings not too high. (We now have bindings for each possible combination.)

Is it right that the count can only appear between an operator and a motion and not anywhere in the middle. So,  ``d2ap`` (delete two paragraphs) is ok, but ``da2p`` is not. (At least in Vi, it doesn't seem to work.) If this is true, that would make the implementation easier. We start in navigation mode and accept both operators, motions and counts. When an operator is given, we go into operator_mode and only accept motions anymore. This also means that ViState will get another field.

I'm excited about this.

yyt16384:
> Is it right that the count can only appear between an operator and a motion and not anywhere in the middle. So, d2ap (delete two paragraphs) is ok, but da2p is not.

Yes, counts can be before the operator or before the motion, and not anywhere else. If there are counts in both places they are multiplied.

jonathanslenders:
Ok! Going to have a look at this very soon. (I had a few other priorities.)

jonathanslenders:
Hi @yyt16384. I merge this in https://github.com/jonathanslenders/python-prompt-toolkit/commit/f56a75a96160b5ee3112ee1603eaf4087a09d092

I reviewed the code, which looks excellent, but I'll give it a while before I push a new release.

But we can already proceed with adding a separate state ``waiting_for_motion`` in ``ViState`` and save the operator over there.

jonathanslenders:
These changes have been merged. Further I did more refactoring of the Vi key bindings, and now we do have separate states and key bindings for text objects and operators.
So closing this. Thanks a lot for your help!

PR #255, opened 2016-02-28
Update documentation on creating full screen applications.

Carreau:
Just re-following the tutorial blindly, and make some update that make some things clearer (IMHO).


jonathanslenders:
Thanks for this. It's merged now.

PR #256, opened 2016-02-28
Carry the weight in layout dimensions to allow stretching

Carreau:
Otherwise things like the following don't work

```python
layout = VSplit([
    
    Window(D(weight=2), ...),

    Window(width=D.exact(1),   content=FillControl('|', token=Token.Line)),

    Window(D(weight=2), ...)),
])
```

### DIFF
--- a/prompt_toolkit/layout/containers.py
+++ b/prompt_toolkit/layout/containers.py
@@ -1001,7 +1001,7 @@ class Window(Container):
         else:
             max_ = dimension.max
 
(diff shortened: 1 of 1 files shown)


jonathanslenders:
Thanks for this. Definitely makes sense. Merged in: https://github.com/jonathanslenders/python-prompt-toolkit/commit/e478c46f06af2ce599c4a8db84ab55d12565ff96

PR #259, opened 2016-02-29
Share code between HSplit, VSplit

Carreau:
H and VSplit share a lot of code,create an intermediary class. 

It seem that beyond the Swap width/height, the only difference is in `extended_height`, I suppose this is to allow things to overflow through the bottom of the screen. Would it make sens to introduce a (potentially useless)  `extended_width` and find a way to share more code ?


Carreau:
Failure seem random, cannot curl smth. Can you relaunch this travis test ?

Issue #252, opened 2016-02-26
Autosuggestion, add optimistic (no)update ?


Carreau:
I got a lag when the autosuggestion take times to build the list of suggestion, 
assuming vertically is time, and `|` is my cursor:

```
imp|ort
impo|
impo|rt
impo|
impor|t
```

You can easily reproduce that with : 

```
from prompt_toolkit import prompt
from prompt_toolkit.history import InMemoryHistory
from prompt_toolkit.auto_suggest import AutoSuggestFromHistory

class MyHist(InMemoryHistory):

    def __iter__(self):
        import time
        time.sleep(0.3)
        return super().__iter__()

history = MyHist()

while True:
    text = prompt('> ', history=history, auto_suggest=AutoSuggestFromHistory())
    print('You said: %s' % text)
```

My guess is that it should be possible to optimistically keep the previous suggestion if the new type letter is the same than the first one of the (previous) suggestions or something similar, or better, if current string is subset of current suggestion, keep it. 

Thoughts?

jonathanslenders:
I think this makes sense. I'll have a look later this week.

Carreau:
No hurry, it seem also pretty easy for a newcomer to investigate the codebase. :-P

PR #261, opened 2016-03-01
Provide simple caching function for autosuggestion

Carreau:
Fix/Mitigate #252 

Also add example. 

I'll cleanup, maybe improve that later. I have to run to a meeting. 
Tell me if you don't think it is the good strategy.

### DIFF
--- a/docs/pages/building_prompts.rst
+++ b/docs/pages/building_prompts.rst
@@ -353,6 +353,11 @@ A suggestion does not have to come from the history. Any implementation of the
 :class:`~prompt_toolkit.auto_suggest.AutoSuggest` abstract base class can be
 passed as an argument.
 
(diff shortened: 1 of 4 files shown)


jonathanslenders:
Thanks! Maybe however I think I will prefer this to be in Buffer class itself. (Not sure at the moment.) Further, we also want to be able to reuse the cache in case of a backspace press.

Carreau:
> edit: after a second thought. I think that maybe this is actually the right solution. Give me some more time.

> we also want to be able to reuse the cache in case of a backspace press.

I though about that too, but it's actually wrong to reuse. 

`matp` complete to `matplotlib` but not `mathematics`. 
Delete one char:
`mat`, might complete to `mathematics` if it's more recent. 

A (LRU ?) cache might be useful for backspace, but we would need a hash of the  history also as part of the key.

PR #286, opened 2016-05-11
Safer list deletion for Registry.remove_binding()

abspoel:
Iterating over a list while modifying it at the same time might lead to
unexpected results:

    >>> lst = list(range(10))
    >>> for x in lst:
    ...     lst.remove(x)
    >>> lst
    [1, 3, 5, 7, 9]

See Lennart Regebro's answer here: http://stackoverflow.com/a/1207427


abspoel:
Never mind, there is actually a return statement in the function I missed, so this is irrelevant.

PR #288, opened 2016-05-12
Fix key bindings for Vi mode registers

yyt16384:
This fixes yanking into named registers, and adds paste from register key bindings only in normal mode.


jonathanslenders:
Thanks a lot @yyt16384!
Merged in: https://github.com/jonathanslenders/python-prompt-toolkit/commit/3cdbd8daff094127b32eff8587fa8d5bbf9da6ab

PR #292, opened 2016-05-19
Ignore common virtualenv locations

tony:
This helps me a lot with playing with your libs (python-prompt-toolkit, pyvim, pymux) in development mode.

It depends on the project, but the common pattern virtualenv locations I see (when making them in the project directory) tend to be:

- ``env/``
- ``venv/``
- ``.venv/``
- ``.env/``
- ``env3/`` (these are esp. helpful since I often hop between a python 2 and python 3 env when testing)
- ``.env3/``
- ``.venv3/``
- ``venv3/``

I've been experimenting and found this regex ignores all of them. If there ever were a directory created with "env" anywhere in the name, this ignore would need to be revised.

What do you think?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,12 @@ lib
 lib64
 __pycache__
 
+# Virtualenvs
+#  Common destinations (depending on developer) are env and venv, they can
+#  sometimes to preceded with a period (e.g. .venv) or followed by a number
+#  (e.g. venv3) or both (.venv3).
+*env*/
+
 # Installer logs
 pip-log.txt
 


jonathanslenders:
Sorry, I forgot to reply here. Are you planning to send pull requests to every Python project to update their gitignore with your preferred virtualenv locations? I appreciate the pull request, but we can't do this. For these kind of things, you can set up a global git ignore. My virtualenv for instance doesn't live in any of these.

tony:
i retract this. it's a distraction from other things

tony:
> Are you planning to send pull requests to every Python project to update their gitignore?

https://github.com/github/gitignore/blob/455a69dd48ce041f6ac2aa7aeeb9560957311e2f/Python.gitignore#L81

So all the gitignore people are a bunch of chumps :smile:? I think its safe to say there are common patterns we want to ignore.

I do stand back at this PR and cringed, not at the idea of ignoring virtualenvs, but the idea I added this ugly glob.

> your preferred virtualenv locations

*My* preferred ones? Now I'm Mr solipsism :smile:? I've used virtualenvwrapper, pyenv-virtualenv, pyenv-virtualenvwrapper... but in a pinch, its very common to just lay down a virtualenv in the project dir. Lots and lots of people are doing it, because we can't depend on knowing what virtualenv helper software someone has, or where they store their virtualenvs normally :P I could give you maybe hundreds of examples of this:

http://flask.pocoo.org/docs/0.10/installation/
http://python-eve.org/install.html

I'm being facetious though. My intention was wholly the *opposite* of my own personal tastes. I want to make life easier for other developer's workflows.

jonathanslenders:
Ok, thank you for explaining! I had no idea that these are common. (I've never seen any of these 8.)
Anyway, I think that virtualenvwrapper is configurable, so I think it's much better to configure locally, system wide anyway.

Also, this would come at the cost/risk of excluding two many files. What if at some point I need a ``prompt_toolkit/layout/env/something/__init__.py``. This is not unusual. So, because of that, I'm a bit conservative here. Hopefully you understand.

Maybe have a look at what the core Python developers are using, as a reference:
https://github.com/python/cpython/blob/master/.gitignore

yyt16384:
Note that you can add a leading slash to match only in root directory. (so `/env/` for example)

tony:
> Note that you can add a leading slash to match only in root directory. (so /env/ for example)

@yyt16384 that's could improve it significantly then  ``/*env*/``.

cpython is the project for a python implementation, so a good place to look general gitignore ideas, but i wouldn't assume it as a representation of core dev's tastes on .gitignore in external python packages. since cpython is a whole implementation, i'm not sure the workflow is the same as other python packages.

using a virtualenv is the best practice, there's no recommendation agreed upon on where to put it. I've even gone as far as trying to create tools to rectify differences in how devs handle virtualenvs: https://github.com/tony/bootstrap_env/blob/master/bootstrap_env.sh. this is part of an effort I've been on to make development workflows for those who fork projects easier.

mypy has it https://github.com/python/mypy/blob/master/.gitignore. Flask does https://github.com/pallets/flask/blob/master/.gitignore. its [very common practice](https://github.com/search?utf8=✓&q=env+%7C%7C+venv+&type=Code) for devs to place virtualenv's in the project dir.  The python guide's example http://docs.python-guide.org/en/latest/dev/virtualenvs/ demonstrates the usage as such.

But, pypa projects like https://github.com/pypa/pip/blob/master/.gitignore, https://github.com/pypa/warehouse do not. django doesn't https://github.com/django/django/blob/master/.gitignore.

maybe i'll write a blog post about it sometime. I don't want to create distraction over this. We can lock this discussion if you don't think its productive @jonathanslenders.

jonathanslenders:
@tony: If you create a pull request that excludes ``/*env*/`` (with the slash), I don't mind merging that.
It looks harmless and it's good if it helps some people. I close the discussion here.

PR #311, opened 2016-06-06
If only one completion available apply it directly.

Carreau:
And do not display the completion menu.

reported in ipython/ipython#9540

### DIFF
--- a/prompt_toolkit/buffer.py
+++ b/prompt_toolkit/buffer.py
@@ -693,6 +693,9 @@ class Buffer(object):
 
         # Set `complete_state`.
         if completions:
+            if len(completions) == 1:
+                self.apply_completion(completions[0])
+                return
             self.complete_state = CompletionState(
                 original_document=self.document,
                 current_completions=completions)


jonathanslenders:
Thanks! Give me a few days to think about it, but I think it definitely makes sense to merge.

Carreau:
I found an edge case if you have a completer bug. 

with this patch, because in IPython `import<tab>` complete only to `importlib`, `im<tab>` with this patch does automatically `im<tab>` ->`import` -> `importlib`without user actions.

jonathanslenders:
Ok, thanks for testing, I'll have a look during the coming days. (I think we have to do this slightly different. There is also the case where ``complete_while_typing`` is true, and completions are generated as the user types, rather than only when 'tab' has been pressed. In this case, we can't automatically do this.)

jonathanslenders:
I made the following change which could possibly fix this:

https://github.com/jonathanslenders/python-prompt-toolkit/commit/1be9ddd77784df236aab140d51994a8aaef161c0

Could you test? I think it does what you want.

A few remarks:
- Notice that you can return completions that don't change anything. For instance in the case of ``import<tab>``, you could return one completion that turns this into ``importlib``, but add a second one that doesn't insert anything and just keeps it as ``import``. This makes it possible to toggle between the completion and go back if the original text was valid. (Of course, doing this will undo the change you were expecting from this pull request. But personally, I prefer it that way.)
- Also notice that completions can change whatever was displayed before the completion. IPython doesn't seem to do case insensitive completion. E.g. ``import`` doesn't complete to ``ImportWarning``. By creating a ``Completion`` instance with a negative start position, you can replace whatever was before the cursor. This is important for users, because they tend to forget the casing of names.

Carreau:
> Could you test? I think it does what you want.

From a quick test, it works. 

> Notice that you can return completions that don't change anything. For instance in the case of import<tab>, you could return one completion that turns this into importlib, but add a second one that doesn't insert anything and just keeps it as import. This makes it possible to toggle between the completion and go back if the original text was valid. (Of course, doing this will undo the change you were expecting from this pull request. But personally, I prefer it that way.)

Yes I was thinking about that, but no need too if you can dismiss the completer, but that's IPython problem, not yours , yo are already doing enough ! :-) 


> Also notice that completions can change whatever was displayed before the completion. IPython doesn't seem to do case insensitive completion. E.g. import doesn't complete to ImportWarning. By creating a Completion instance with a negative start position, you can replace whatever was before the cursor. This is important for users, because they tend to forget the casing of names.

We do do it as `matpl` can completer to `%matplotlib`, sure we don't deal with casing, but we have plan to do that by integrating with Jedi. 

so +1 to this patch !

jonathanslenders:
With the patch that I applied last, I can close this, right?

Carreau:
Yes ! Thanks !

PR #315, opened 2016-06-09
history_search parameter seem to not exist.

Carreau:
Though I'm curious how I can search backward through history without searching :-) 

(requested on IPython ML)

### DIFF
--- a/prompt_toolkit/buffer.py
+++ b/prompt_toolkit/buffer.py
@@ -791,7 +791,6 @@ class Buffer(object):
         Move forwards through the history.
 
         :param count: Amount of items to move forward.
-        :param history_search: When True, filter history using ``self.history_search_text``.
         """
         self._set_history_search()
 


jonathanslenders:
Thanks! This is a remainder of older code. It is possible to have history-search like can be configured for readline. If you type "imp" followed by arrow up, it will walk through the history, but only showing entries starting with "imp". If you want, this can be configured by passing ``enable_history_search=True`` to the ``prompt/create_prompt_application`` function.

jonathanslenders:
Merged in: https://github.com/jonathanslenders/python-prompt-toolkit/commit/412b1730f709825a7e55cd0e8bf4db9aa9547fde

Carreau:
> Though I'm curious how I can search backward through history without searching :-)

Yeah, I was more wondering if 2 different shortcut could have 1 searching and the other not. 
But that was just a question, don't think there is a real usecase. 

Thanks

PR #318, opened 2016-06-11
[WIP] Win32 paste (+pasting when bracketed paste is disabled.)

jonathanslenders:
This branch should fix the situation when pasting multiple lines results in bad handling of the line endings.

TODO:
- handle win32_asyncio event loop
- handle posix + posix_asyncio event loops.

Breaking (internal) change: ``CommandLineInterface._InterfaceEventLoopCallbacks.feed_key()`` can raise ``ComandLineInterfaceIsDoneError`` when trying to feed keys.

### DIFF
--- a/prompt_toolkit/eventloop/win32.py
+++ b/prompt_toolkit/eventloop/win32.py
@@ -6,7 +6,6 @@ Windows notes:
 """
 from __future__ import unicode_literals
 
-from ..prompt_toolkit.interface import CommandLineInterfaceIsDoneError
 from ..terminal.win32_input import ConsoleInputReader
 from ..win32_types import SECURITY_ATTRIBUTES
 from .base import EventLoop, INPUT_TIMEOUT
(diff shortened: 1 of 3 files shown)


stonebig:
thanks, will test in 2 hours

jonathanslenders:
For me, I just noticed that this doesn't solve the indentation issue yet.

For ptpython, people can press ``F6`` and then it enables paste mode, which disables both auto-ident and accept-on-enter.

Maybe for Windows, we can also use ``PeekConsoleInputW`` and if we see that the pending input contains either more than 16 characters, or contains a newline (for instance) consider it a bracketed paste.
(going to try this tomorrow.)

stonebig:
the multi-line copy/paste works \o/

jonathanslenders:
Hi @stonebig,

Can you test again with the following branch instead?
https://github.com/jonathanslenders/python-prompt-toolkit/pull/319

This emulates bracketed paste on Windows. It means that if text is pasted on Windows, no additional indentation will be added, and all input is kept in one input field.

stonebig:
done. so far so good, now going to test the other ptipython fix

jonathanslenders:
Thanks for testing. Closing this branch in favor of: https://github.com/jonathanslenders/python-prompt-toolkit/pull/319

PR #319, opened 2016-06-12
Win32 paste2

jonathanslenders:


jonathanslenders:
Merged. See this option: https://github.com/jonathanslenders/python-prompt-toolkit/blob/master/prompt_toolkit/shortcuts.py#L85

PR #325, opened 2016-06-17
Add cursor_position property to Completion

koljonen:
`cursor_position` says how far from the end of the completion the cursor should be placed, so that we can have completions such as e.g. `<html[cursor]` -> `<html>[cursor]</html>`.

### DIFF
--- a/examples/autocompletion-with-cursor-position.py
+++ b/examples/autocompletion-with-cursor-position.py
@@ -1,0 +1,31 @@
+#!/usr/bin/env python
+"""
+Autocompletion example with cursor position not at the end of the completion.
+
+Press [Tab] to complete the current word.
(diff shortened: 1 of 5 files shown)


jonathanslenders:
Thanks @koljonen for this.

This is definitely useful. I think there are a few situations that are not correctly handled in the pull request which I'll have to check later on. And also, I'm not entirely sure that I want to expand ``WordCompleter`` (In a way, I'd like to keep ``WordCompleter`` to be an example of a completer that is almost as simple as possible.)
Right now, I have some other priorities, but I come back to this issue.

koljonen:
I've squashed away the changes to WordCompleter.

lwm:
Can you rebase @koljonen?

PR #328, opened 2016-06-24
Do not re-invoke completer if only 1 completion.

Carreau:
closes ipython/ipython#9658

### DIFF
--- a/prompt_toolkit/interface.py
+++ b/prompt_toolkit/interface.py
@@ -855,22 +855,22 @@ class CommandLineInterface(object):
 
                         set_completions = True
                         select_first_anyway = False
-
-                        # When the commond part has to be inserted, and there
+                        # When the common part has to be inserted, and there
                         # is a common part.
                         if insert_common_part:
                             common_part = get_common_complete_suffix(document, completions)
                             if common_part:
                                 # Insert + run completer again.
                                 buffer.insert_text(common_part)
-                                async_completer()
+                                if len(completions) > 1:
+                                    async_completer()
                                 set_completions = False
                             else:
                                 # When we were asked to insert the "common"
                                 # prefix, but there was no common suffix but
                                 # still exactly one match, then select the
                                 # first. (It could be that we have a completion
-                                # which does * expension, like '*.py', with
+                                # which does * expansion, like '*.py', with
                                 # exactly one match.)
                                 if len(completions) == 1:
                                     select_first_anyway = True


Carreau:
I'm trying to see how to test that, but I seem to have issues as a cli should run with an event loop. The autocompletion animal example seem to have the same issue if one animal ends with space.

jonathanslenders:
This looks good. Going to test/merge asap. Thank you!

If you're interested, in order to write a unit test, it could be done similar to this: https://github.com/jonathanslenders/python-prompt-toolkit/blob/master/tests/test_cli.py#L49

Carreau:
Ha ! Thanks ! I'll have a look at a test !

Carreau:
Humm, I cam up with something like that, which work in the REPL

```
+completer = WordCompleter([
+    'space ',
+    'nospace']
+, ignore_case=True)
+
 def _feed_cli_with_input(text, editing_mode=EditingMode.EMACS):
     """
     Create a CommandLineInterface, feed it with the given user input and return
@@ -32,6 +39,7 @@ def _feed_cli_with_input(text, editing_mode=EditingMode.EMACS):
             eventloop=loop,
             input=inp,
             output=DummyOutput())
+        cli.buffers[DEFAULT_BUFFER].completer=completer
         result = cli.run()
         return result, cli
     finally:
@@ -39,6 +47,14 @@ def _feed_cli_with_input(text, editing_mode=EditingMode.EMACS):
         inp.close()


+def test_simple_no_pick_completion():
+    # Simple text input, followed by enter.
+    result, cli = _feed_cli_with_input('spac\x09\n')
+    assert result.text == 'space ' , 'got %s'%result.text
+
+test_simple_no_pick_completion()
+
+
 def test_simple_text_input():
     # Simple text input, followed by enter.
     result, cli = _feed_cli_with_input('hello\n')
```

unfortunately it seem like pipe input does not trigger completions when fed with \x09 or \t.

jonathanslenders:
The reason it doesn't work is probably because the completer runs asynchronously. If something else is pressed in the meantime before the completions are generated, then those completions are discarded.

We would need to sleep after pressing tab. But I'm not sure whether that plays well with the way we use the event loop in the tests right now. (I'd like to make it more similar to asyncio, so that we can write better tests. Don't worry too much if it doesn't work right now.)

jonathanslenders:
Hi @Carreau,
Just to keep you posted. I have had very little time during the past few days. Also tomorrow and probably the day after tomorrow, I won't have much time. But I hope to finish merging this and a few other small fixes + performance improvements before the end of next week, doing a new release then.

Jonathan

Carreau:
No problem, I have a workaround in IPython.

jonathanslenders:
This should be fixed in: https://github.com/jonathanslenders/python-prompt-toolkit/commit/4f653f32f1516610689c481f301f54b78d3d242a
(I fixed it slightly different.)

Carreau:
Thanks, I'll try that with IPython workaround removed !

[still need to ship tshirt now that I'm back from SciPy]

PR #333, opened 2016-06-30
lazy pkg_resources import

scopatz:
This implements a background `pkg_resources` import when using pygments.

Hopefully, some of the other lazy tools here can also be used to speed up prompt-toolkit startup as well.

### DIFF
--- a/prompt_toolkit/layout/lexers.py
+++ b/prompt_toolkit/layout/lexers.py
@@ -191,6 +191,10 @@ class PygmentsLexer(Lexer):
         """
         Create a `Lexer` from a filename.
         """
+        # must come before pygments imports
+        from ..lazyasd import load_module_in_background
+        load_module_in_background('pkg_resources', debug='PROMPT_TOOLKIT_DEBUG',
+            replacements={'pygments.plugin': 'pkg_resources'})
         from pygments.util import ClassNotFound
         from pygments.lexers import get_lexer_for_filename
 
--- a/prompt_toolkit/lazyasd.py
+++ b/prompt_toolkit/lazyasd.py
@@ -1,0 +1,280 @@
+"""Lazy and self destrctive containers for speeding up module import."""
+# Copyright 2015-2016, the xonsh developers. All rights reserved.
+import os
+import sys
+import time
+import types
+import builtins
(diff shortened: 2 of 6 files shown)


jonathanslenders:
Hi @scopatz,

This is a change that I prefer not to include in prompt-toolkit itself. One reason is that many users are import pygments themselves, and we can't be sure that our imports are the first one. Further, it's an ugly workaround for something that has to be fixed in Pygments itself.

Is it correct that calling ``load_module_in_background`` outside of prompt-toolkit, but at the very beginning of the application would be sufficient to get the speedup?

btw, the latest master branch should now have a few other speedups in the key bindings that you could appreciate.

Jonathan

scopatz:
> One reason is that many users are import pygments themselves, and we can't be sure that our imports are the first one. 

Sure, in which case this code will simply fall back to the previous import.

> Further, it's an ugly workaround for something that has to be fixed in Pygments itself.

I agree that the fix is deeper down the stack.  If not pygments, it should be in pkg_resources or setuptools.  I don't have a lot of confidence that this will be fixed ever though, which is what led me to this solution.  I think the benefits to the users offset the cost of a relatively small amount of ugliness.

> Is it correct that calling load_module_in_background outside of prompt-toolkit, but at the very beginning of the application would be sufficient to get the speedup?

Yes it is.  I just figured that this also affecting all other prompt-toolkit project and thus helps perpetuate the myth that Python has to be slow.  It is totally fine if you don't feel this is appropriate; was just a suggestion. 

I would appreciate it, though, if other people hit this slow import issue, if you would point them to this solution.  I am planning on splitting lazyasd into its own package, so that other people can use it without xonsh.

> btw, the latest master branch should now have a few other speedups in the key bindings that you could appreciate.

Awesome!

jonathanslenders:
I'm closing this. Prompt_toolkit 2.0 doesn't require Pygments anymore. So, by default it won't import pkg_resources at all during start up.
Let me know if I'm missing something that would still require something like this for Xonsh.

scopatz:
Thanks for pushing through with this.

scopatz:
Glad that pygments is no longer a requirement.

PR #337, opened 2016-07-11
Minor fix to save a redirect

shabda:
RTD links now use a .io domain, .org will redirect but .io is the newer recommendation.


jonathanslenders:
Thanks! Merged in: https://github.com/jonathanslenders/python-prompt-toolkit/commit/c7e84aa45fde0da61a4ecefeaaf96aca61d79d17

PR #344, opened 2016-07-15
Allow `words` to be a callable for WordCompleter

kevinjqiu:
so a hook can be called to fetch the word list under different context

### DIFF
--- a/prompt_toolkit/contrib/completers/base.py
+++ b/prompt_toolkit/contrib/completers/base.py
@@ -12,7 +12,7 @@ class WordCompleter(Completer):
     """
     Simple autocompletion on a list of words.
 
-    :param words: List of words.
+    :param words: List of words, or a callable that produces a list of words.
     :param ignore_case: If True, case-insensitive completion.
     :param meta_dict: Optional dict mapping words to their meta-information.
     :param WORD: When True, use WORD characters.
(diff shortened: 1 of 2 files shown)


kevinjqiu:
@jonathanslenders Any chance this feature could be mainlined? I modified this to support fetching of auto-complete list based on the current context (e.g., the current selected database) used in this project of mine: https://github.com/kevinjqiu/cdbcli

The calling code is here: https://github.com/kevinjqiu/cdbcli/blob/master/cdbcli/completer.py#L50-L53

If prompt-toolkit already supports this feature through other means, let me know and I will update my code to use that. Otherwise, please consider mainline this as I do not want to vendor prompt-toolkit inside my project.

Thanks!

jonathanslenders:
Hi @kevinjqiu,

Sorry for replying this late, I have spend most of my time lately with a few bugs.

Anyway, I could merge this. (after some changes -- I'll comment in the pull request.) However, keep in mind that prompt_toolkit is meant to be extensible. This means that if somebody needs a custom completer, he can create a custom ``Completer`` class and implement that interface. There's no need to vendor prompt_toolkit, just for this.

But I think I'll merge this. Also, cdbcli looks like a nice project!

prompt_toolkit/contrib/completers/base.py
```diff
@@ -12,7 +12,7 @@ class WordCompleter(Completer):
     """
     Simple autocompletion on a list of words.
 
-    :param words: List of words.
+    :param words: List of words, or a callable that produces a list of words.
```
jonathanslenders: For consistency with some other parts of prompt_toolkit, I think it's better to have a second parameter named ``get_words``.

kevinjqiu:
@jonathanslenders No problems, and thanks for the reply.

I've addressed your feedback here https://github.com/jonathanslenders/python-prompt-toolkit/pull/344/commits/62ba67fa3eedd31f2d9aa457ea683ed80c822376

The reason I didn't create a custom completer is that in my use case, `WordCompleter` is sufficient after the minor change to make the fetching of word list context-aware, which I feel is a nice addition to the `WordCompleter` class.

Thanks for creating prompt-toolkit! It's an awesome library.

kevinjqiu:
@jonathanslenders Any updates? I have addressed your code review feedback.

jonathanslenders:
This has been merged here (with some changes): https://github.com/jonathanslenders/python-prompt-toolkit/commit/edeb20bebba4bc8d2932654d046a1253cc218b78

Sorry it took that long!

I kept ``words`` as an argument instead of ``get_words``, because that's more consistent with the rest of prompt_toolkit 2.0.

Thanks again for proposing this change!

PR #348, opened 2016-07-16
Support -ve argument for C-k. Fixes #347

### LINKED ISSUE #347: Negative argument to C-k should kill text backwards
The normal behavior of `C-k` is to kill text to the end of the line. But with a -ve argument, it changes to kill text to the beginning of the line ([see](https://www.gnu.org/software/bash/manual/html_node/Readline-Arguments.html)). Right now, it behaves just like `C-k`.

NOTE: I have a patch that is working and will send it in shortly.

haridsv:


jonathanslenders:
Hi @haridsv,

Thanks a lot for this. I merged it here: https://github.com/jonathanslenders/python-prompt-toolkit/commit/c5b07a4228b5b4cb20953d02d608ea8cd8f71c38
Also added a unit test: https://github.com/jonathanslenders/python-prompt-toolkit/commit/be68caacffec4695379fd241903689e5e6731688

haridsv:
Thank you, and thank you for adding a unit test, I meant to figure that out, but it slipped my mind.

PR #351, opened 2016-07-19
Failsafe for running from the python terminal

orbweaver-:
If you try to run the application from idle and make a call to the interface, it gives a very cryptic error. This will at least make it easier to understand that you can't do that.


jonathanslenders:
Hi @orbweaver-,

Thanks for the patch! However, I don't understand how we can ever get a ``TypeError`` from the code in the try/except. Can ``os.environ.get('TERM')`` raise an exception? Or am I missing something?

jonathanslenders:
If you don't mind, I'm going to close this. (Catching ``TypeError`` is not correct, given that this cannot happen.) Feel free to reopen this pull request if I'm missing something.

Thanks again for the help!

Issue #367, opened 2016-08-04
Can't run gevent-get-input.py on Windows (ptpython and IPython have the same issue)


maryush:
Hi, i install ptpython from pip, but when i try to start i got error like:

```
Traceback (most recent call last):
  File "c:\python27\lib\runpy.py", line 162, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "c:\python27\lib\runpy.py", line 72, in _run_code
    exec code in run_globals
  File "c:\python27\lib\site-packages\ptpython\__main__.py", line 7, in <module>
    run()
  File "c:\python27\lib\site-packages\ptpython\entry_points\run_ptpython.py", line 71, in run
    title='Python REPL (ptpython)')
  File "c:\python27\lib\site-packages\ptpython\repl.py", line 309, in embed
    cli.run()
  File "c:\python27\lib\site-packages\prompt_toolkit\interface.py", line 389, in run
    self.eventloop.run(self.input, self.create_eventloop_callbacks())
  File "c:\python27\lib\site-packages\prompt_toolkit\eventloop\win32.py", line 66, in run
    self._inputhook_context.call_inputhook(ready)
  File "c:\python27\lib\site-packages\prompt_toolkit\eventloop\inputhook.py", line 82, in call_inputhook
    select_fds([self._r], timeout=None)
  File "c:\python27\lib\site-packages\prompt_toolkit\eventloop\select.py", line 49, in select_fds
    result = _select(read_fds, timeout)
  File "c:\python27\lib\site-packages\prompt_toolkit\eventloop\select.py", line 99, in _select
    return select.select(read_fds, [], [], timeout)[0]
select.error: (10038, 'Pr\xf3bowano przeprowadzi\xe6 pr\xf3b\xea na obiekcie, kt\xf3ry nie jest gniazdem')
```

I spend some time to tested it and found that one prompt toolkit example not working to with the same errors:

```
c:\tmp\examples>gevent-get-input.py
Give me some input:
Traceback (most recent call last):
  File "C:\tmp\examples\gevent-get-input.py", line 21, in <module>
    answer = prompt('Give me some input: ', eventloop=eventloop)
  File "c:\Python27\lib\site-packages\prompt_toolkit\shortcuts.py", line 540, in prompt
    eventloop=eventloop)
  File "c:\Python27\lib\site-packages\prompt_toolkit\shortcuts.py", line 621, in run_application
[...]

PR #368, opened 2016-08-04
Don't call select() on Windows in InputHookContext

takluyver:
Windows select() can't be used on fds, so this will always fail.

I've no idea if the gevent problem this was meant to prevent can occur on Windows.

Closes gh-367

See also ipython/ipython#9836

### DIFF
--- a/prompt_toolkit/eventloop/inputhook.py
+++ b/prompt_toolkit/eventloop/inputhook.py
@@ -79,7 +79,9 @@ class InputHookContext(object):
             # monkey patched and won't be cooperative, so that would block all
             # other select() calls otherwise.
             # See: http://www.gevent.org/gevent.os.html
-            select_fds([self._r], timeout=None)
+            if os.name != 'nt':
+                # Windows select() can't be used on fds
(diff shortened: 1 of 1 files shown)


jonathanslenders:
Yes, this is the fix. Thanks! See: https://github.com/jonathanslenders/python-prompt-toolkit/commit/6d30d445d49c5c8c39df9f66ba40cda3e259edeb
I did some testing and the gevent bug that this should handle does indeed not occur on Windows.
It's weird that select.select did not cause any issue on my Windows system.
Anyway, I'll push a new release now.

takluyver:
If you tested on Python 3, I think that would have obscured the bug: due to a reorganisation of exceptions, the error it raises now inherits from OSError, which is caught in `call_inputhook()`. All four reports of this issue came from Python 2.7 users.

takluyver:
Thanks for pushing a new release out so quickly :)

jonathanslenders:
That explains. Even on Python 3 that would have been a bug. It would essentially cause the inputhook to quit immediately. This was bad :(

PR #374, opened 2016-08-11
SentenceValidator: Remove set comprehension to support python 2.6

joguSD:
Set comprehensions aren't available in python 2.6, so I've converted it to a list comprehension cast as a set.


jonathanslenders:
Thank you! Merged in: https://github.com/jonathanslenders/python-prompt-toolkit/commit/56ee5ac8225ffb7320c2b640160d0860dba3dee6

PR #382, opened 2016-08-17
Suppress exceptions coming from signals while running the prompt_toolkit eventloop.

jonathanslenders:
Instead, quit the eventloop, and raise these exceptions afterwards.


takluyver:
Presumably it will mess things up if inputhooks set signal handlers? I don't think any of ours do directly, but I don't know what e.g. Qt is doing when you start its event loop. There's probably not much we can do about that, though.

prompt_toolkit/interface.py
```diff
@@ -1159,3 +1166,73 @@ def add_reader(self, fd, callback):
...
+        # receive signals anyway.)
+        if isinstance(threading.current_thread(), threading._MainThread):
+            # Replace each signal handler.
+            for n in self._get_signal_numbers():
+                try:
+                    original_handler = signal.getsignal(n)
+                except ValueError:  # Signal number out of range.
+                    pass
+                else:
+                    if original_handler:
```
takluyver: This will be True for ignored signals: `SIG_IGN` is like 1, but is not callable. We probably don't want to modify ignored signals, and possibly not signals with the 'default action' (which means killing the program for most signals).

takluyver: That wasn't very clear. The return value of `getsignal()` is a function *or* one of the special values `SIG_IGN` or `SIG_DFL`. Those are not callable.

jdemeyer:
I have not tested this code, but it seems to make one assumption which is not true in general: that all OS-level signal handlers are set by Python.

Python's `signal()` actually does two things: it sets the OS-level signal handler to some internal handler and it sets the Python-level signal handler to some Python callable.

In some cases (e.g. using [cysignals](https://github.com/sagemath/cysignals) or other libraries), the OS-level signal handler is a custom handler and that would not be reset correctly. In other words, `signal(SIG, getsignal(SIG))` is not guaranteed to be a no-op.

Unfortunately, from pure Python code, these internals cannot be accessed. Ideally, it would be possible to change the Python-level signal handler without changing the OS-level handler (say, a function `setsignal` as counterpart to `getsignal`). But you need to patch Python for that...

jonathanslenders:
@jdemeyer: that's a very good point. This won't work in these cases, and could possibly do more bad than good.

So, that would mean that technically, it's just impossible to find a correct solution what what we were trying to do. I can still think of reading the input in another thread (because they are free from signals), but also that could break other things. (input hooks for instance).

I suggest that we merge https://github.com/ipython/ipython/pull/9867/files after putting those changes in a finally block. It's still not 100% correct, but it won't become worse.
Just keep in mind that registering signal handlers that raise an exception will always be dangerous and fragile, unless we have control of all the code that runs in your process. Nothing to do about that.

PR #385, opened 2016-08-21
Avoid extra vertical scrolling on windows if the cursor is already on screen

stephenrauch:
```win32_output.scroll_buffer_to_prompt()``` tries to scroll the cursor onto the screen with a preference to place the cursor at the bottom if the buffer is larger than the view-port.  This does not take in to account that the cursor may already be in the view-port.  If the cursor is in the view-port, it is moved to the bottom, and then very shortly there after moved up 8 lines to account for the reserved area for the menu.  Using https://github.com/click-contrib/click-repl this causes the screen to flicker anytime the screen is scrolled.

This PR leaves the vertical scroll unchanged if the cursor is already in the view-port.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,3 +37,6 @@ nosetests.xml
 
 # Generated documentation
 docs/_build
+
+# pycharm metadata
+.idea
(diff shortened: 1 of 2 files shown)


jonathanslenders:
Hi @stephenrauch,

Thank you for this. I just tested it and it looks good!
Merged in: https://github.com/jonathanslenders/python-prompt-toolkit/commit/192e8e6690a05f01a519fd1bd310cc6853fa00e4

PR #388, opened 2016-08-31
ensure get('TERM') returns a string

minrk:
otherwise `'xterm' in None` will raise TypeError when TERM is not in env

cf ipython/ipython#9921


jonathanslenders:
Sorry, I just notice this pull request now, after having done exactly the same fix myself: https://github.com/jonathanslenders/python-prompt-toolkit/commit/99fa7fae61c9b4ed9767ead3b4f9b1318cfa875d

Thanks anyway for taking the time to solve this!

minrk:
No problem!

PR #396, opened 2016-10-01
win32 console debug

stephenrauch:
Debugging applications on windows under a debugger/ide can be difficult
because the IDE may use its own 'console' which Prompt Toolkit is unable
to use productively.  This PR makes it possible to set an environment
variable which allows Prompt Toolkit to create a console window and
attach the appropriate handles to allow the program to use the console
productively.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -9,3 +9,4 @@ Contributors
 ------------
 
 - Amjith Ramanujam <amjith.r AT gmail.com>
+- Stephen Rauch <stephen AT rauch.net>
--- a/docs/pages/building_prompts.rst
+++ b/docs/pages/building_prompts.rst
@@ -677,3 +677,72 @@ loop. Prompting the user for input is as simple as calling
 The ``patch_stdout=True`` parameter is optional, but it's recommended, because
 other coroutines could print to stdout. This option ensures that other output
 won't destroy the prompt.
+
+
+Debugging on Windows
+--------------------
+
+Debugging applications on windows under a debugger/ide can be difficult
+because the IDE may use its own 'console' which Prompt Toolkit is unable
+to use productively.  In these cases it is possible to set an environment
+variable which allows Prompt Toolkit to create a console window and
+attach the appropriate handles to allow the program to use the console
+productively.
+
+Two conditions need to be present for this to work. First, the code needs
+to be imported before anything else that may use the std file handles, or
+the console, because they may cache these handles, and thus wouldn't work
+as hoped.
+
+.. code:: python
+
+    from prompt_toolkit.utils import is_windows
+    if is_windows():
+        from prompt_toolkit.terminal import win32_console_debug
+
+        # check if any environment variables are set that induce us to action
+        win32_console_debug.check_environment_variables()
+
+The second condition needed is to set an environment variable::
+
+   SET USE_WIN_CONSOLE=Window Identifier
+
+or alternatively, if managing the import directly, right after the import,
+call:
+
+.. code:: python
+
+    win32_console_debug.ensure_console(title="Window Identifier")
+
+or, if it is preferred to only invoke when the application is being
+'debugged', set an environment variable::
+
+   SET USE_WIN_CONSOLE_DEBUG=Window Identifier
+
+or right after the import, call:
+
+.. code:: python
+
+    win32_console_debug.ensure_console(
+        debugger_only=True, title="Window Identifier")
+
+The console will be reused (based on the "Window Identifier" passed in)
+when restarting the application.
+
+
+Other Console Support (or starting in pre-existing terminals)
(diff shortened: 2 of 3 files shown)


jonathanslenders:
Hi @stephenrauch,
Thank you for this! It looks like you spend on lot of time on this patch.
I need a bit of time to read through all of the code before merging it, and probably I'll have some remarks.
But it sounds definitely useful!
Jonathan

jonathanslenders:
Hi @stephenrauch,

I just review the code. In general it looks good. But I'd like a few changes. (Hopefully you don't mind.)

- Most importantly, importing ``win32_console_debug`` has a side effect. It executes ``_check_environment_variables``. We should not execute any code during an import. It's better to call this function ourself. (See: https://chrismorgan.info/blog/say-no-to-import-side-effects-in-python.html )
- Further, I prefer to avoid introducing new environment variables. I would prefer that users who need this functionality, simple call a function. Something like ``create_win32_console(debug=False)``. If they like to expose this API through an environment variable, then they can still do it themself.

If you don't have the time to fix any of this, then I'll do it.
I just tested this change in my Windows machine with Pycharm, and it works wonderful!

Thanks,
Jonathan

stephenrauch:
I will gladly make the changes.  Give me a few days.

Cheers,

-- Stephen

stephenrauch:
@jonathanslenders 

So I have finally updated the pull request.  I wanted to use it for a while to help make sure it was working ok.  I am glad I did, as it needed a ^C handler.  There are now no side effects of the import.  I did however leave the Environment Variable as an option for invoking the functionality, for this reason:  Editing the source code of my program between actually running it in the console, and running it under the debugger is rather a PITA.  This way I can assign the environment variable in my debug environment, and it is largely a NOP when not debugging.  The use of the environment variable is optional, as discussed in the docs, the functionality can be invoked without it.

Hopefully this meets with your desires.  If not, or you need anything else, please LMK.

I have also added (a needed) ^C handler.

pe224:
@stephenrauch 
Could you give me a hint on how to get your console to work?
I invoke it by
```
from prompt_toolkit.terminal.win32_console_debug import ensure_console
ensure_console()
```
On the invoking side, I get
```
Found debug console pid 5468,  pipe name: 'FIND_CONSOLE: ipy_repl.py'`
```
A console window opens up which outputs
```
debug window pipe name: \\.\pipe\FIND_CONSOLE: ipy_repl.py
Connected at 2017-07-02 22:24:27.539
Error in sys.excepthook:

Original exception was:
```
So there is a connection followed by an exception.
The exception seems to occur when `h_pipe = win32pipe.ConnectNamedPipe(pipe)` is called the 2nd time in the final while-loop of your stub program.
I'm on Python 3.6, 64bit.

stephenrauch:
@pe224 So I am reading right that there is no exception information printed?  Also, what windows version?

pe224:
@stephenrauch Correct. It's Windows 10.

jonathanslenders:
Hi @stephenrauch,

Excuse me for coming back just now.

Regarding this pull request, I'm not sure what to do. I still think it's definitely valuable, but I'm not sure I have the resources and knowledge to maintain this piece as part of prompt_toolkit while it is evolving. I also think that it's perfectly feasible to implement all of this on top of prompt_toolkit, like it is right now. What about creating a prompt_toolkit_win32_debugger package or something like that?
Probably this requires also a couple of changes in order to be compatible with the 2.0 release.

Spectrem12:
This pull request was never merged. Is there another solution for debugging while using an IDE?

jonathanslenders:
@Spectrem12,

PyCharm has now a built-in console emulator. It's not enable by default, but you can do this in the settings:

image

I'm not sure, but VS Code probably has a similar option. Since Windows now has ConPTY (see https://devblogs.microsoft.com/commandline/windows-command-line-introducing-the-windows-pseudo-console-conpty/ ), any IDE can render a console.

So, I don't think there's a need for this anymore.

@stephenrauch : hope you're fine too if we close this?

Spectrem12:
I should have added more detail to my initial question.

I have the emulator option checked but it limits the debugging and output capabilities or maybe I just don't know about pycharm.

All the output is monochrome. The output buffer is limited. And you are not able to type inline python syntax within the emulator.

I appreciate any solutions or work around you may have to offer.

jonathanslenders:
I'm not sure about the debugging capabilities, because something is broken in my PyCharm right now, but colors are definitely working fine (notice I run it from the PyCharm run button, it's not the "terminal" pane).

image

Spectrem12:
Yes, I am also using the run button.

Are those colors defined within the IDE itself or with an extension?

I am using Grep Console and it no longer functions using the emulated window (not exactly prompt-toolkit issue) , but it causes other setbacks such as being able to pause debugging and input python commands inline.
As well, as I am not able to have PyCharm open up my separate log files that have streams going to with Grep Console working on those either.

Is there a workaround to not force the terminal emulator?

PR #398, opened 2016-10-03
Fix the "colors" example in the "Building prompts" chapter

xiaq:


xiaq:
I also noticed that the examples mix use of `prompt` from both the `prompt_toolkit` module and `prompt_toolkit.shortcuts`. Do you think it makes sense to consistently import from `prompt_toolkit.shortcuts`?

jonathanslenders:
Thank you! These commits have been merged.

Regarding importing ``shortcuts``. I'm not sure. I think it makes sense for people to know that this is an abstraction that builds the UI for a simple prompt and that it comes from the shortcuts module. However, I will also never remove the import in ``prompt_toolkit/__init__.py``. So, I don't have much opinion about this.

PR #400, opened 2016-10-04
CI & test tweaks:

mindw:


jonathanslenders:
Thank you! Going to review/merge this asap.

mindw:
ping?

jonathanslenders:
@mindw There are a lot of changes in this pull request and not all of them are related. Could you split this in multiple pull requests, that each do just one thing? I'm also not sure which changes are actually fixing something, and which changes are rather a "cleanup".
It makes it quite hard to do a code review.
So, it would be great if you can make a separate commit and pull request for each independent change, and then make sure that the commit message explains why the change is required. Thanks!

PR #403, opened 2016-10-08
Multi-line insertion

davidbrochart:
Added Shift-I Vi command in block selection mode to insert text at the beginning of each line of the block.

### DIFF
--- a/prompt_toolkit/key_binding/bindings/vi.py
+++ b/prompt_toolkit/key_binding/bindings/vi.py
@@ -4,7 +4,7 @@ from prompt_toolkit.buffer import ClipboardData, indent, unindent, reshape_text
 from prompt_toolkit.document import Document
 from prompt_toolkit.enums import IncrementalSearchDirection, SEARCH_BUFFER, SYSTEM_BUFFER
 from prompt_toolkit.filters import Filter, Condition, HasArg, Always, to_cli_filter, IsReadOnly
-from prompt_toolkit.filters.cli import ViNavigationMode, ViInsertMode, ViReplaceMode, ViSelectionMode, ViWaitingForTextObjectMode, ViDigraphMode, ViMode
+from prompt_toolkit.filters.cli import ViNavigationMode, ViInsertMode, ViInsertSelectionMode, ViReplaceMode, ViSelectionMode, ViWaitingForTextObjectMode, ViDigraphMode, ViMode
 from prompt_toolkit.key_binding.digraphs import DIGRAPHS
 from prompt_toolkit.key_binding.vi_state import CharacterFind, InputMode
 from prompt_toolkit.keys import Keys
@@ -161,6 +161,7 @@ def load_vi_bindings(registry, enable_visual_key=Always(),
     #  ViState says different.)
     navigation_mode = ViNavigationMode()
     insert_mode = ViInsertMode()
+    insert_selection_mode = ViInsertSelectionMode()
     replace_mode = ViReplaceMode()
     selection_mode = ViSelectionMode()
     operator_given = ViWaitingForTextObjectMode()
@@ -407,6 +408,18 @@ def load_vi_bindings(registry, enable_visual_key=Always(),
         event.cli.vi_state.input_mode = InputMode.INSERT
         event.current_buffer.cursor_position += \
             event.current_buffer.document.get_start_of_line_position(after_whitespace=True)
+
+    @handle('I', filter=selection_mode & ~IsReadOnly())
+    def _(event):
+        event.current_buffer.selection_cursor_positions = []
+        for i, from_to in enumerate(event.current_buffer.document.selection_ranges()):
+            event.current_buffer.selection_cursor_positions.append(from_to[0])
+            if i == 0:
+                event.current_buffer.cursor_position = from_to[0]
+        event.cli.vi_state.input_mode = InputMode.INSERT_SELECTION
+        buffer = event.current_buffer
+        if bool(buffer.selection_state):
+            buffer.exit_selection()
 
     @handle('J', filter=navigation_mode & ~IsReadOnly())
     def _(event):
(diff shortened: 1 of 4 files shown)


jonathanslenders:
Wonderful! Thank you for this!

I just tested it locally, and it works beautifully. I'll have a second look later today or tomorrow and then I'll merge it.

davidbrochart:
Thanks!
I just realized backspace and delete keys don't work though. Any idea?

jonathanslenders:
Yes, I noticed. But that's a following step. I'd like to merge that separately, first I'd like to be sure that this is absolutely correct.
(There are a few other keys in this mode that don't work, like ControlW for instance.)

davidbrochart:
I just committed a fix for the backspace/delete handling.

davidbrochart:
I just committed the same thing for Shift-A (insert at the end of the selection block).

prompt_toolkit/key_binding/bindings/vi.py
```diff
@@ -1464,6 +1490,60 @@ def _(event):
...
+    def _(event):
+        """
+        Insert data at the end of each line of the block selection.
+        """
+        cursor_position = event.current_buffer.cursor_position
+        selection_right_cursor_positions = event.current_buffer.selection_right_cursor_positions
+        insert = False
+        if event.data == '\x7f': # Backspace
+            j = [-1, -1, -1]
+        elif event.data == '\x1b[3~': # Delete
```
jonathanslenders: Checking ``event.data`` is not reliable. (This won't work on Windows or when people configure a terminal to send different escape sequences. Better is to have separate key bindings, using for instance ``Keys.Delete``.

prompt_toolkit/key_binding/vi_state.py
```diff
@@ -9,6 +9,8 @@
 
 class InputMode(object):
     INSERT = 'vi-insert'
+    INSERT_SELECTION_LEFT = 'vi-insert-selection-left'
+    INSERT_SELECTION_RIGHT = 'vi-insert-selection-right'
```
jonathanslenders: I'm not sure that we really need both ``INSERT_SELECTION_LEFT`` and ``INSERT_SELECTION_RIGHT``.

prompt_toolkit/key_binding/bindings/vi.py
```diff
@@ -408,6 +410,30 @@ def _(event):
...
+    @handle('A', filter=selection_mode & ~IsReadOnly())
+    def _(event):
+        event.current_buffer.selection_right_cursor_positions = []
+        for i, from_to in enumerate(event.current_buffer.document.selection_ranges()):
+            event.current_buffer.selection_right_cursor_positions.append(from_to[1] + 1)
+            if i == 0:
+                event.current_buffer.cursor_position = from_to[1] + 1
+        event.cli.vi_state.input_mode = InputMode.INSERT_SELECTION_RIGHT
+        buffer = event.current_buffer
+        if bool(buffer.selection_state):
```
jonathanslenders: Do we need the cast to ``bool``?

jonathanslenders:
FYI: I left a few comments, but I'm about to start merging, with some changes.

prompt_toolkit/key_binding/bindings/vi.py
```diff
@@ -1464,6 +1490,60 @@ def _(event):
...
+            else:
+                event.current_buffer.delete()
+            for i2 in range(i1 + 1, len(selection_left_cursor_positions)):
+                selection_left_cursor_positions[i2] += j[1]
+        event.current_buffer.cursor_position = cursor_position + j[2]
+
+    @handle(Keys.Backspace, filter=insert_selection_right_mode)
+    @handle(Keys.Delete, filter=insert_selection_right_mode)
+    @handle(Keys.Any, filter=insert_selection_right_mode)
+    def _(event):
```
jonathanslenders: The code between inserting left and right is identical. I'm going to replace this by a feature I'd call "multiple cursors". Whether the insert positions are coming from the left or right side of the selection doesn't matter. We just have multiple positions where we can insert text.

davidbrochart:
My code was probably not following good practices, so thanks a lot for your comments. Looking forward to seeing this feature in next release!

jonathanslenders:
Don't worry! The pull request saved me some work, so it's was valuable anyway.
I thing I probably also have to do is create a unit test for this. (Not much work.)

jonathanslenders:
@davidbrochart, Thanks again for this! I merged your first commit, and made some changes for the rest. Normally, everything should be in prompt_toolkit.

davidbrochart:
Thank you for merging and for the enhancements!

davidbrochart:
This last commit is (not exactly) an implementation of the Vi substitute command (search and replace). I just took advantage of the multi-line insertion commit to go a bit further. In selection mode, it is now possible to enter the search mode (with `/`) and then to enter the replace mode (with a second `/`). Once the replace mode is entered, the matching text (in the selection) is deleted, and the user can enter the replacing text. The replace mode is exited by pressing `Esc`.
So it doesn't follow the Vi `:s/foo/bar/g` syntax, but at the same time it still seems quite logical: you start with a search and you go on with a replace if you want. One limitation is that now the `/` character cannot be used in the searched text. It is just a proof of concept, let me know if you find it interesting/acceptable (and sorry for using `global` variables!).

davidbrochart:
Or, instead of the second `/`, we could just hit `Enter`, just like in pure search mode. This way, there wouldn't be any limitation of not having a `/` character in the searched text. And because replacing is active only in selection mode, this behavior doesn't interfere with the pure search mode. So, the sequence would be:
`/` -> enter searched text -> `Enter` -> enter replacing text -> `Esc`

jonathanslenders:
Hi @davidbrochart,

About searching in selection mode, this has been implemented, but slightly different. In commits: https://github.com/jonathanslenders/python-prompt-toolkit/commit/b2171606e60a5e1d60cc355b57e9611c18e86e8e
and https://github.com/jonathanslenders/python-prompt-toolkit/commit/e84abde00ffb6039de58e4aa555acbee81c766af

About the search/replace mode. I'm not a big fan of using enter to go into replace mode. It doesn't work like that in Vi or other editors, right?

Just one question, if you'd like to propose more improvements (which are always very welcome), could you start from a new pull request? That makes it easier to review changes (and I could close this one). Thanks!

davidbrochart:
Sure, I will create a new pull request, so that you can close this one.
I know it is not the Vi syntax, I will try to implement the real substitute command then.

PR #405, opened 2016-10-15
Implemented change repetition (using '.') in VIM mode.

zardus:
Hello,

Thank you for this incredible library. Since ipython switched to prompt_toolkit, I've been doing more and more actual development without leaving ipython. I pretty quickly noticed that vim's "repeat last change" functionality didn't make it into prompt_toolkit's vi mode. This PR implements it.

To accomplish this, I save off the relevant functions (handlers, operators, and text object creators) and event objects directly. I'm not sure if storing event objects like this is legit, but it makes it simple to repeat changes and operations and to recreate text objects. An alternative is to store the data necessary to recreate the events, but I'm not familiar enough with the code to accomplish this yet.


zardus:
I should also mention that this change doesn't *fully* support the semantics of `.` in vim. Specifically, `.` can be used to repeat actions taken in insert mode (including changes using the `c` and `s` operators), but this change doesn't support it. If there's an easy way to get to the previous inserted text, that wouldn't be a hard thing to support. Is getting at the inserted text possible?

Also, I noticed that, when I try to repeat a replacement (`r`), the replaced letter is always ".". It looks like `event.data` is overwritten on the *saved* event object. Interestingly, `event.arg` does not suffer from this. Any ideas? I could save off event.data specially, but at that point, it sounds like it might be cleaner to save off the necessary info to recreate the event itself.

jonathanslenders:
Hi @zardus,

Thanks a lot for this! It's definitely useful.
So far, I did not had any time yet to review the changes, but I will.
(It could be that I have to change a few things, but in any case, I very much appreciate the help.)

Jonathan

zardus:
I rebased this on the current master, as the text object decorator code had been moved around.

PR #336, opened 2016-07-10
Implement Emacs Meta-. key binding (insert last word)

mbachry:
Now when IPython uses prompt_toolkit, I miss readline's `Meta-.` keybinding a lot. Here's my attempt to implement it.

PR #407, opened 2016-10-20
Added Emacs mode Meta-. binding 'yank last arg'

### LINKED ISSUE #343: Implement readline's `yank-last-arg`
The implementation of readline's `yank-last-arg` command [is currently a TODO](https://github.com/jonathanslenders/python-prompt-toolkit/blob/master/prompt_toolkit/key_binding/bindings/emacs.py#L298). It would be good for it to actually get implemented!

adamchainz:
Fixes #343. Alternative implementation to #336 that uses a single function and also implements `yank-nth-arg`, in the same style as readline.

### DIFF
--- a/prompt_toolkit/key_binding/bindings/emacs.py
+++ b/prompt_toolkit/key_binding/bindings/emacs.py
@@ -1,5 +1,7 @@
 # pylint: disable=function-redefined
 from __future__ import unicode_literals
+import re
+
 from prompt_toolkit.buffer import SelectionType, indent, unindent
 from prompt_toolkit.keys import Keys
 from prompt_toolkit.enums import IncrementalSearchDirection, SEARCH_BUFFER, SYSTEM_BUFFER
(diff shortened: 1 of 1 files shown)


jonathanslenders:
Hi @adamchainz,

Cool! Thanks a lot for this.
I still need to take some time to compare this implementation and the other.
Probably, I'd like to go for a solution where the key bindings are completely stateless, and the state is stored in the ``Buffer`` object. (Need to think about this.)

Jonathan

adamchainz:
Happy to change that but I thought it was rather elegant that it can all be done with function-bound 'static' variables 😄

jonathanslenders:
Hi @adamchainz,

I combined both pull request in the following commit: https://github.com/jonathanslenders/python-prompt-toolkit/commit/0c09bf034aeeb2facbfbc070cc4fa2f58620baa5

If you have time, could you have a look and see whether this implements all you need?

Thanks,
Jonathan

adamchainz:
Thanks! Sure, will do this evening.

PR #409, opened 2016-10-27
Import prompt_async shortcut in __init__, too

iamale:
So the example in https://python-prompt-toolkit.readthedocs.io/en/stable/pages/building_prompts.html#prompt-in-an-asyncio-application would work. (Also, it's (probably) more convenient? :^)


jonathanslenders:
Hi @iamale, this commit has been merged in https://github.com/jonathanslenders/python-prompt-toolkit/commit/5e85bf5a790e0c108f036031e3a66dcf163cc899.

Thanks for this!

PR #412, opened 2016-10-29
Implementation of a search-and-replace feature

davidbrochart:
Hi @jonathanslenders,

Here is a search-and-replace feature that is a first step towards the Vi substitute command. It doesn't follow the `:s/foo/bar/g` syntax, instead you need to hit `s` in selection mode, enter the searched text, hit `/`, enter the replacing text, and finally hit `Esc`. The matching text inside the selection will be replaced.
I couldn't implement the Vi syntax because I don't know how `:s` can be catched by prompt-toolkit instead of pyvim (I used pyvim to test this feature), or if it should be catched in pyvim. If you have any idea, please let me know.

David.

### DIFF
--- a/prompt_toolkit/buffer.py
+++ b/prompt_toolkit/buffer.py
@@ -261,6 +261,8 @@ class Buffer(object):
         # using multiple cursors.)
         self.multiple_cursor_positions = []
 
+        self.substitute_selection_ranges = []
+
         # When doing consecutive up/down movements, prefer to stay at this column.
         self.preferred_column = None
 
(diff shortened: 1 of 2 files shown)


davidbrochart:
When we get the exact Vim syntax, we can have neovim's incremental substitute for free ;-)
https://asciinema.org/a/92207

jonathanslenders:
Yeah, that's something I'd like to have for Pyvim at some point. However, it will probably never be fully compatible. Having the performance of Vi and its regex engine is not possible in pure Python.

For this in particular, I'm also considering to go for the "multiple cursors" route. It should be possible to turn a search operation in multiple cursors. But then again, that's not something standard Vi, and I prefer to keep following standards.

PR #428, opened 2016-12-05
Swap vi mode forward and backward search, like readline

altaurog:
I understand the reasoning behind the change in behavior, but IMO backwards search in REPL history is a much more common use case than search within a multiline edit.

All this commit does is switch them.  If you want to get fancy, you could make it configurable.

### DIFF
--- a/prompt_toolkit/key_binding/bindings/vi.py
+++ b/prompt_toolkit/key_binding/bindings/vi.py
@@ -1747,8 +1747,8 @@ def load_vi_search_bindings(registry, get_search_state=None,
     selection_mode = ViSelectionMode()
     handle = create_handle_decorator(registry, filter & ViMode())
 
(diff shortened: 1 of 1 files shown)


altaurog:
This addresses issue #431

jonathanslenders:
Hi @altaurog, Thanks for the pull request. However, I can't fix it yet, because it would break pyvim, pymux and pypager, all applications that rely on the original ordering.
However, It's important to me that this gets fixed, so I'll look for an option somewhere.

altaurog:
Yah, I realized that.  Let me know if you want my help.

jonathanslenders:
Hi @altaurog, this has been implemented in: https://github.com/jonathanslenders/python-prompt-toolkit/commit/9a79ca172891cf03995d75e044f5663ee3c10e17
(slightly different from your pull request.)
In any case, thanks a lot for taking the time to write this pull request!

PR #452, opened 2017-01-27
Fix IndexError in operate-and-get-next command

qweeze:
Ctrl+O shortcut is broken when `buffer.working_index` is pointing at the last working line.


jonathanslenders:
Thanks! Merged in https://github.com/jonathanslenders/python-prompt-toolkit/commit/6ed2477b9bea4aae017bdccc6aa69020dc900c33
(Probably I change a few things in the future, because we should avoid accessing private attributes.)

PR #462, opened 2017-02-04
Handle arguments in the $EDITOR or $VISUAL env variable

Carreau:
On OSX, users will typically use `subl -w` to use sublime tet on
blocking mode, though this will be passed ass it to subprocess.call
leading to the `subl -w` command not found.

See ipython/ipython#9929

this should fix that. It would affect users with actual space on the
path of their editor executable, but I'm unsure how to handle both case.
It seem to me like passing flags would be the most common use case.

Workaround is to tell users to create an shim executable which pass the
options.

### DIFF
--- a/prompt_toolkit/buffer.py
+++ b/prompt_toolkit/buffer.py
@@ -1316,7 +1316,7 @@ class Buffer(object):
         for e in editors:
             if e:
                 try:
-                    returncode = subprocess.call([e, filename])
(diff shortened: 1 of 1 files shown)


jonathanslenders:
Merged in: https://github.com/jonathanslenders/python-prompt-toolkit/commit/8b7d360f4f8c9dc47b87c339e05129b1333e12d0

Thank you @Carreau!

PR #465, opened 2017-02-07
fallback to default terminal size if reported as 0

davidszotten:


jonathanslenders:
Hi David, thank you for this!
Merged in: https://github.com/jonathanslenders/python-prompt-toolkit/commit/e73b9212991d84aed56d004223ead01d8156a369

PR #467, opened 2017-02-13
Fix raw input on SmartOS

gmcmanus:
I found that on SmartOS (Illumos/Solaris derivative OS), IPython would not echo my input until I had typed 4 characters. I tracked it down to how prompt-toolkit was setting the terminal attributes: it was not setting VMIN (min characters to read). On SmartOS this defaults to 4. On Linux this defaults to 1. Setting it to 1 fixed my problem.

It may be worth checking this behaves correctly on Mac OS X - there is a comment that says tty.setraw fails for this OS, and tty.setraw also sets VMIN = 1. Can it be confirmed exactly how tty.setraw fails for Mac OS X? It may be worth just using tty.setraw instead.

### DIFF
--- a/prompt_toolkit/terminal/vt100_input.py
+++ b/prompt_toolkit/terminal/vt100_input.py
@@ -459,6 +459,8 @@ class raw_mode(object):
         else:
             newattr[tty.LFLAG] = self._patch_lflag(newattr[tty.LFLAG])
             newattr[tty.IFLAG] = self._patch_iflag(newattr[tty.IFLAG])
+            newattr[tty.CC][termios.VMIN] = 1
+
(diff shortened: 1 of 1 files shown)


jonathanslenders:
Hi @hmcmanus, Thanks a lot for this. I still need to test this on Mac and Linux, but if it works fine, then it can be merged.

hmcmanus:
@gmcmanus FYI.

jonathanslenders:
Hi @gmcmanus: Thanks for this. I tested on Mac and Linux and everything seems to work fine. Merged in: https://github.com/jonathanslenders/python-prompt-toolkit/commit/613b83d2afbbb2d0b89cdd3fe7772e1a6bfddd32

PR #481, opened 2017-03-02
use matching backticks in docstrings.

Carreau:
We autogenerate a shortcut list in IPython docs using docstrings, the non-matching backticks make RST to complain. 

We could fix it, but it seem cleaner to do it upstream :-)


jonathanslenders:
Thank you for this! I'll merge it ASAP. It makes sense to fix it here. Don't hesitate to create pull requests for stuff like this.

jonathanslenders:
Merged in https://github.com/jonathanslenders/python-prompt-toolkit/commit/021e6cc963570e1bc645defcdeb2da10d0c83a9e + also pushed to the 2.0 branch.

Thanks!

PR #484, opened 2017-03-08
Don't save undo state with CPRResponse

### LINKED ISSUE #419: Unable to implement redo in ipython 5.0.0
I tried implementing a redo keybinding but failed.
I'm not sure why, but it seems like the _redo_stack member is being reset, even though I pass `save_before=lambda x: False` when calling `add_keybinding` on the `key_bindings_registry`.

Code I'm running in ipython:
```python
from IPython import get_ipython
from prompt_toolkit.keys import Keys

def undo(event):
  event.current_buffer.undo()
  # This line prints a non-empty redo stack
  print 'redo stack: {}'.format(event.current_buffer._redo_stack)

# Bind Alt+_ to undo
get_ipython().pt_cli.application.key_bindings_registry.add_binding(Keys.Escape, u'_', filter=True, save_before=lambda x: False)(undo)
```

After running that code, I typed something in the prompt and then invoked `Alt+_` to trigger the undo, which printed a non-empty redo stack.
However, the redo stack of all buffers look empty. To see this I'm running the following:

```python
for b in get_ipython().pt_cli.application.buffers.values():
  print b._redo_stack
```

asmeurer:
This gets called all the time, and ends up clearing the redo state instantly
after an undo.

Fixes #419.

### DIFF
--- a/prompt_toolkit/key_binding/bindings/basic.py
+++ b/prompt_toolkit/key_binding/bindings/basic.py
@@ -196,7 +196,7 @@ def load_basic_bindings():
         """
         event.current_buffer.insert_text(event.data)
 
-    @handle(Keys.CPRResponse)
+    @handle(Keys.CPRResponse, save_before=lambda e: False)
     def _(event):
         """
         Handle incoming Cursor-Position-Request response.


jonathanslenders:
Thanks! Very good catch!

jonathanslenders:
Merged in: https://github.com/jonathanslenders/python-prompt-toolkit/commit/bb8db8e4318afc07ad29b8d59bc4eedc764cd9bc

Issue #482, opened 2017-03-06
CPR responses appearing as text


bmerry:
I'm writing a netcat-like tool using prompt-toolkit, and I'm finding that when a chunk of traffic arrives (leading to a lots of quick calls to `print_tokens` via `run_in_terminal` that I sometimes see "^[[24;1R" repeated multiple times on the terminal. From what I can tell, this is the VT100 reply to a position request.

I could be barking up the wrong tree, but I suspect that the "with self.input.cooked_mode():" in run_in_terminal may be responsible: presumably at some earlier point the renderer asked for position, but the input was switched into cooked mode before the reply was received. If I remove that line, then my immediate problem goes away; but I don't know if that's a good place to fix this or if it's just hiding a symptom of a more fundamental sequencing problem.

For reference, the app I'm developing is uploaded at https://github.com/bmerry/ntsh.

jonathanslenders:
Hi @bmerry,

It's not going to be the ``cooked_mode``. This is a timing issue. The CPR response arrives after the input is done reading. Sometimes this happens in a prompt_toolkit application if you hold Enter for some time. (Which would draw many prompts after each other.)

I was working on a fix for this. (https://github.com/jonathanslenders/python-prompt-toolkit/pull/339/files ) This would wait for a CPR response anytime, but the delay (even small) was not acceptable.

It's a difficult issue to tackle, given the asynchronous nature between sending the CPR request (through stdout) and receiving the response (through stdin). Further, we can't rely on even getting a response.

What I probably can do is adding an option to disable CPR. The result is that the prompt_toolkit application can't know the distance between the prompt and the bottom of the terminal, but for many applications, that is sufficient.

I'll keep this issue open as bug/feature request.

bmerry:
I had a look at that pull request. It looks like it's dealing with the case where the prompt loop is exiting, and I can see why there is a problem there. In my case the prompt is still running, and I'm asynchronously (with asyncio) printing above the prompt. So my assumption was that prompt_toolkit should still be able to deal with CPR responses.

jonathanslenders:
Even in that case, I believe that this can happen, unfortunately.

bmerry:
When disabling cooked mode (#487) I see no more CPR responses on-screen while the prompt is running.

jonathanslenders:
Does this still happen? Some refactoring to the CPR handling has been done in prompt_toolkit 2.0.
If not, I'd like to close this issue.

[...]

PR #487, opened 2017-03-12
Add run_in_terminal option to disable cooked mode

bmerry:
Adds a `cooked_mode` optional argument to run_in_terminal. Using it
avoids CPR responses being echoed to the terminal (see #482). It is not
a complete solution because they can still appear once the prompt exits
and thus leaves raw mode, but avoids some instances.

The default is left at the old behaviour for backwards compatibility.

### DIFF
--- a/prompt_toolkit/interface.py
+++ b/prompt_toolkit/interface.py
@@ -611,7 +611,7 @@ class CommandLineInterface(object):
         if self.eventloop:
             self.eventloop.stop()
 
-    def run_in_terminal(self, func, render_cli_done=False):
+    def run_in_terminal(self, func, render_cli_done=False, cooked_mode=True):
         """
         Run function on the terminal above the prompt.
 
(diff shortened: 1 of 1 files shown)


jonathanslenders:
Hi @bmerry,

Thanks for this pull request. Actually, I still have no idea how this possibly influences CPRs.
I need to think about this and do some more testing to really see what's going on, because I think that somewhere we face a timing issue somewhere. (I'm not sure that this behaviour will be the same on all systems.)

randy3k:
I agree with @jonathanslenders that it should not influences CPRs. However, I think it is still a very good feature to have, see my issue #501.

@bmerry, by switching to `raw_mode`, there is actually the escaped sequence `'\x1b[6n'` queuing in stdin. I guess they just disappear when you are processing them.

[1]: https://github.com/jonathanslenders/python-prompt-toolkit/blob/e73b9212991d84aed56d004223ead01d8156a369/prompt_toolkit/terminal/vt100_output.py#L626

jonathanslenders:
Ok, thanks for this. I agree it's useful enough to merge.

jonathanslenders:
Merge in: https://github.com/jonathanslenders/python-prompt-toolkit/commit/70795a7a13044515ebe1b2d28e0993f447fbf5e5 Thanks everyone!

Issue #411, opened 2016-10-28
Problems understanding PipeInput


fspegni:
Hi. I'm studying your library since a while, but didn't find a proper way of fixing my issue.

In brief: I have a customized app, eventloop, and cli. I want the user to pass a custom input (let's say a `PipeInput`) to the cli.

I send 4 `\n` separated lines on the `PipeInput`, but it looks to me the prompt doesn't process three commands.

A minimal script is:
```
import prompt_toolkit as pt


def my_prompt(prompt_in):
    """
    Inspired by prompt_toolkit.shortcuts.prompt()
    """

    line = None

    try:
        app = pt.shortcuts.create_prompt_application(
            erase_when_done=True,
        )
        el = pt.shortcuts.create_eventloop()
        cli = pt.interface.CommandLineInterface(
            application=app,
            input=prompt_in,
            eventloop=el,
            output=pt.shortcuts.create_output(true_color=False)
        )
        patch_context = pt.utils.DummyContext()

        with patch_context:
            doc = cli.run(reset_current_buffer=True)

            if doc:
                line = doc.text.strip()

    except EOFError:
        line = None

    return line

# setup the input
prompt_in = pt.input.PipeInput()
prompt_in.send("foo\nfie\nfoo\nfie\n")

# start the interpretation loop

while True:

    try:

        line = my_prompt(prompt_in)

        if line is None:
            break

        print "[%s]" % line
    except KeyboardInterrupt:
        print "Goodbye"
        break
    except Exception, e:
        print "Error: %s" % e
        break

```

The output is:
```
user@host:wd$ python example.py  
foo
[foo]
^[[23;1R^[[23;9R
```

I can't understand what those strange sequences of characters mean, and how to debug this problem.

Hope you can shed some light :) Thanks

fspegni:
I forgot to specify: as you can see the interpretation stops at the first command (foo), instead of continuing with the next

jonathanslenders:
Hi @fspegni,

``PipeInput`` is actually mostly meant for unit-testing. You could use it to redirect input, for instance in the case of an SSH or Telnet connection, where no TTY was allocated, but usually, you'd read from stdin. What is the actual use case you're trying to implement?

Regarding the escape characters that you see. They are the CPR response. Cursor-position-request-response. In order to know how much space there is available below the prompt, we send a CPR request to stdout. The terminal will reply with the cursor coordinates through stdin. You would have to read these characters from stdin.

[...]

PR #488, opened 2017-03-13
handle the case of non-terminal input and output

fspegni:
In this commit I handled a use case that is apparently not covered by your code (as discussed in #411). 

Roughly speaking: the same prompt can be configured to read from a file/stream or from a terminal; in the first case the output goes to a file/stream (tipically the stdout) without caring of the cursor position (there is no need of sending CPRs). In the second case (prompt reading from a terminal), the Vt100_Output class does its job just fine.

I did two main things:
- slighly changed the renderer to consume one char at the time, so if reading from a file a don't mix the input of two different lines
- added a StandardOutput class that does most of the job that Vt100_Output did, but it ignores the CPRs

At the moment StandardOutput is only used for linux systems, but I don't have a windows system to test with. We can consider making StandardOutput a common class for both the linux terminal output (Vt100_Output) and the windows one (I don't know what class handles it).

### DIFF
--- a/prompt_toolkit/eventloop/posix_utils.py
+++ b/prompt_toolkit/eventloop/posix_utils.py
@@ -49,11 +49,18 @@ class PosixStdinReader(object):
         #: True when there is nothing anymore to read.
         self.closed = False
 
-    def read(self, count=1024):
+    def read(self, count=1):
             # By default we choose a rather small chunk size, because reading
             # big amounts of input at once, causes the event loop to process
             # all these key bindings also at once without going back to the
             # loop. This will make the application feel unresponsive.
+            #
+            # Update: if count > 1 and you read from a file (e.g. a script) 
+            # you are going to skip newlines and mix two successive commands;
+            # usually a prompt input is not a huge stream, so reading it 1-by-1
+            # is not too bad, but perhaps there's a better way to fix the above
+            # problem -FS 
+    
         """
         Read the input and return it as a string.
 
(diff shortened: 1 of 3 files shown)


jonathanslenders:
Hi @fspegni,

Thanks a lot for this pull request. I really appreciate it.

However a few issues/comments:
- The reason that count was 1024, is because this makes pasting text much, much faster. Reading one character at a time would make a single paste really slow. This is mostly an issue when pasting big chunks of text. I guess that I should have documented this better.
- The output algorithm can change any time. We intend to support vt100 terminals and Windows consoles as output device. However, using the same algorithm for writing to a file or piping to a different process doesn't guarantee a consistent or meaningful output. Turning this into a feature would make it impossible to create more rendering optimizations in the future.
- This has been discussed before (I forgot where exactly), but prompt_toolkit is only meant for user interaction. It's not going to be a machine-interaction interface. If you want to make a program accept input from a pipe, or send output to a pipe or file, then it's better to just check ``sys.stdout.isatty()`` and if so, just read the input. Thinking about separation of concerns, it's always better to separate the code that does user interaction from the code that contains the business logic.
- Further, I'm working on a prompt_toolkit 2.0 version, which refactors the input and output classes. I don't intend to merge new features to the current master branch, unless there's a big need for them.
- For the future, don't comment old code in a pull request. If something is intended to be deleted, really delete it. This makes the diff much easier to read and review.

So, unfortunately I can't merge this pull request, but again, thanks for the help!
Jonathan

fspegni:
Hi @jonathanslenders ,

thanks for your feedback. First of all, sorry for badly commenting the code in my pull request: I'm such a novice in remote collaboration using pull requests, and should have avoided that. Thanks for your patience :)

I guess my only option then is to handle 2 different interfaces, one for the interactive CLI, and a different one for processing scripts. Not a big deal.

Thanks

PR #495, opened 2017-03-29
Remove a duplicate checking of row range.

c-bata:
This is not a serious issue.

I noticed `get_cursor_up_position` method has duplicate checking of the row range because `translate_row_col_to_index` method is:
https://github.com/jonathanslenders/python-prompt-toolkit/blob/master/prompt_toolkit/document.py#L308-L310


c-bata:
Oh, sorry. this patch has a bug... I'll close.

PR #506, opened 2017-05-16
BUG: Use yield from in Win32AsyncioEventLoop

eric-wieser:
The previous code was not correct, as `yield from` also pushes exceptions to the iterator, which this did not.

This pattern is already used elsewhere

### DIFF
--- a/prompt_toolkit/eventloop/asyncio_win32.py
+++ b/prompt_toolkit/eventloop/asyncio_win32.py
@@ -25,13 +25,15 @@ class Win32AsyncioEventLoop(EventLoop):
         self.closed = False
         self.loop = loop or asyncio.get_event_loop()
 
+    # The following `run_async` function is compiled at runtime
+    # because it contains syntax which is not supported on older Python
+    # versions (yield from)
+    six.exec_(textwrap.dedent('''
     @asyncio.coroutine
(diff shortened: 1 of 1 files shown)


jonathanslenders:
Hi @eric-wieser,

Thank you for the pull request. The ``exec_`` is not going to do a lot without catching a ``SyntaxError``.

This is one of the things that should have been fixed on the 2.0 branch already. I consider that the way forward. Honestly, I prefer not to do big changes anymore on this branch, unless important or obvious bug fixes. Especially the event loop is easy to get wrong, and actually there are more asyncio related bugs on this 1.0 branch. Hopefully that's fine.

eric-wieser:
> The exec_ is not going to do a lot without catching a SyntaxError.

Oops, you're right - I forgot this was in a class suite, not a function one.

I was not aware of the 2.0.0 branch.

One thing I'm puzzled by - this file has a `import asyncio` at the global level. Does that not mean that it can use `yield from` directly, as it would `ImportError` anyway?

eric-wieser:
(And you're correct, this is fixed in the 2.0 tree)

jonathanslenders:
Assuming I can close this. Let me know if that's not the case.

PR #520, opened 2017-07-01
Enable autowrap in renderer again (for possible background threads.)

jonathanslenders:


ngg:
+1 for this, bg thread linewrap now works correctly from ipython

Issue #532, opened 2017-08-12
History search matching anywhere


ahundt:
I believe the current history search matches from the start of the string. Could it be updated with an option for matching anywhere in the string?

The `enable_history_search` could perhaps be updated where the current `True` and `False` settings do as they have done, but now there are additional options for `enable_history_search=`:

- `'starts-with'` (same as current `enable_history_search=True` setting) 
- `'contains'` (matches anywhere in the string).

lwm:
Hi @ahundt, from https://github.com/jonathanslenders/python-prompt-toolkit/pull/534#issuecomment-322538767, I can see that the idea of the change should go in, which is cool :+1: 

I am not sure about your changes in https://github.com/ahundt/python-prompt-toolkit/commit/e62d188ac311695192b3fb1495f8e5d4ddd619df. It seems like the purpose of `enable_history_search` is to enable the functionality, not to configure the functionality. What do you think about adding a different optional argument, perhaps a callable (which might reduce having to specify explicitly all the keywords) that configures the searching? The default would then be the existing functionality.

This seems like a better API to me. Just some thoughts from a passing issue gardener :)

ahundt:
I'll defer to you on design. I think python prompt toolkit 2 has a completely different design so this PR may no longer be useful. 

Please note that this is a code base I'm not familiar with, so I just tried to implement it with the fewest possible modifications.

ahundt:
@jonathanslenders Is this now possible to do in 2.0.x? If so, could you explain how & then this issue can likely be closed.

PR #534, opened 2017-08-13
history search if line contains or starts with string

ahundt:
(#532)

### DIFF
--- a/prompt_toolkit/buffer.py
+++ b/prompt_toolkit/buffer.py
@@ -209,11 +209,13 @@ class Buffer(object):
     :param complete_while_typing: :class:`~prompt_toolkit.filters.SimpleFilter`
         instance. Decide whether or not to do asynchronous autocompleting while
         typing.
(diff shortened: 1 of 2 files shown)


ahundt:
Those filter classes like SimpleFilter seems to only support true and false options or I'd have used it. Hopefully my approach is reasonable but I'm open to make changes.

jonathanslenders:
Hi @ahundt,

Thanks a lot for this pull request.
The 1.0 branch is actually bugfix only, but I'd like to merge this in the 2.0 branch once I've had the time to review it.

It's very useful.

ahundt:
cool, thanks!

ahundt:
@jonathanslenders I've been using the existing functionality a bit more and I realized along with this change I'd also love for the selection of completion to be case-insensitive. In other words if there is a `LICENSE.txt` in a folder and you type `lic<tab>`, `LICENSE.txt` would be an option. I believe that would also be a very small change much like this one.

PR #555, opened 2017-09-22
2.0: add on_resize callback

randy3k:
Added a resize callback. It would be useful for some applications which need to control the width of output print. For example in [rice](https://github.com/randy3k/rice/issues/20), this callback is used to change the output print width.


randy3k:
Just realized that it doesn't work with Windows. I may need to poll the terminal width regular.

PR #561, opened 2017-11-11
2.0: fix search toolbar text for multiline mode

### LINKED ISSUE #553: 2.0 multiline search toolbar shows wrong text
```
        p = Prompt()
        p.prompt("> ", multiline=True)
```

Hit `ctrl-r` and the search tool shows `I-search` instead of `I-search Backward`.

randy3k:
fix #553 

The cause is the `app.layout.current_control` being the search control instead of the default control. So `app.current_search_state` fails to get the correct state. This fix allows the search control to access the search state of the default buffer.


randy3k:
Cc @jonathanslenders 

Bump!

jonathanslenders:
Thanks you @randy3k!

Sorry for replying late.

PR #562, opened 2017-12-03
Add pypy3 to travis testcases.

traverseda:


traverseda:
Well, that was a bit more involved than I thought it would be, but there it is. Failing tests *seems* to be failing on their own merit.

return42:
If you really want to test py26 & py33 you have to stick the pytest version. See pytest's [changelog](https://docs.pytest.org/en/latest/changelog.html#deprecations-and-removals). But take into account that both are no longer supported, see release notes [3.3](https://www.python.org/download/releases/3.3.0/) and [2.6](https://www.python.org/download/releases/2.6.9/)

traverseda:
I do not, I was just going off of what was already there.

jonathanslenders:
Hi all, on the 2.0 branch, I already added pypy3. No new features will be added to the 1.0 (master) branch, so not sure whether this is worth merging.

traverseda:
Ahh, that's fair. But if I might suggest, it's a good idea to have the major development reflected in master. Consider making 2.0 master and 1.0 the branch. Makes the development status a lot clearer at a glance.

jonathanslenders:
Thanks for the suggestion! I made 2.0 the default in GitHub. Probably, I'll switch the branch names later on.

PR #574, opened 2018-01-17
2.0: renamed default_style to default_ui_style.

j-bennet:
Fixed some full-screen examples (`default_style` does not exist anymore).


jonathanslenders:
Hi @j-bennet, thanks for noticing this!
Actually, the right fix would be to not include ``default_ui_style`` at all. It is already included in any application by default. So, if you want, you can also remove the ``merge_styles`` function in these examples.

j-bennet:
@jonathanslenders Good to know! I'll change that.

j-bennet:
@jonathanslenders Changed.

jonathanslenders:
Thanks!

jonathanslenders:
Thanks!

PR #576, opened 2018-01-22
bugfix in flush key at 2.0 branch

### LINKED ISSUE #570: cursor position reset after scrolling screening when using 2.0 branch
I have been using prompt-toolkit to develop a console for the R language: https://github.com/randy3k/Rice

Recently, I was told that the cursor position is automatically reset when a user tries to scroll screen via a terminal emulator: https://github.com/randy3k/rice/issues/31

The following is a screenshot of the demonstration of the issue
jan-04-2018 00-59-52
The console was running on a [terminal emulator](https://github.com/Wramberg/TerminalView) of Sublime Text which uses [pyte](https://github.com/selectel/pyte) behind the scene. You can see that the cursor position is reset when I tried to scroll up the screen. 

After some analyses, I have determined that the problematic code is related to a recent update 
https://github.com/jonathanslenders/python-prompt-toolkit/blob/2.0/prompt_toolkit/key_binding/key_processor.py#L309

If that line is commented, the scrolling works as expected and the cursor position does not reset.
jan-04-2018 01-06-00

My guess is that prompt-toolkit somehow messes up the scroll up and down sequences `<ESC>D` and `<ESC>M`.

randy3k:
1. only invalidate ui when some keys are processed
2. do not start_timeout if the last key is _Flush, otherwise it causes a dead loop
3. only send flush key when the queue is not empty

close #570

### DIFF
--- a/prompt_toolkit/key_binding/key_processor.py
+++ b/prompt_toolkit/key_binding/key_processor.py
@@ -283,11 +283,19 @@ class KeyProcessor(object):
             else:
                 return self.input_queue.popleft()
 
+        keys_processed = False
+        is_flush = False
+
         while not_empty():
+            keys_processed = True
+
             # Process next key.
             key_press = get_next()
 
-            if key_press.key != Keys.CPRResponse:
+            is_flush = key_press is _Flush
+            is_cpr = key_press.key == Keys.CPRResponse
+
+            if not is_flush and not is_cpr:
                 self.before_key_press.fire()
 
             try:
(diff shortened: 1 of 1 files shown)


jonathanslenders:
Thank you @randy3k!
This was a serious bug. I really appreciate it!
I'll review and merge tomorrow.

prompt_toolkit/key_binding/key_processor.py
```diff
@@ -387,7 +398,7 @@ def wait():
             " Wait for timeout. "
             time.sleep(self.timeout)
 
-            if counter == self._keys_pressed:
```
randy3k: @jonathanslenders I think that it should check `key_buffer` instead of `input queue`, right?

jonathanslenders:
That's right. Thanks @randy3k 
Fixed here: https://github.com/jonathanslenders/python-prompt-toolkit/commit/0a649798a42e09e1583c9f84183ec4bc8202d5c2

PR #608, opened 2018-04-01
removed confusing (not required) style code and fixed bottom_toolbar error

pythonpsql:
The removed code was not necessary


pythonpsql:
assert in line 217 in style.py failed when 'class:bottom-toolbar' was used instead of only 'bottom-toolbar'

jonathanslenders:
@pythonpsql, thanks a lot for this merge request, for all the fixes! Really appreciate it.

PR #611, opened 2018-04-10
reset completion state as a fallback

randy3k:
I figured out that in some situations, the `tab` key for requesting completions is not working because previous completion state is not clear. This PR clears the state if none of the conditions were matched.

### DIFF
--- a/prompt_toolkit/buffer.py
+++ b/prompt_toolkit/buffer.py
@@ -1508,6 +1508,10 @@ class Buffer(object):
                         # exactly one match.)
                         if len(completions) == 1:
                             self.go_to_completion(0)
+                else:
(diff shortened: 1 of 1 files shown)


jonathanslenders:
Hi @randy3k, thanks for this!
It's merged, but I made a small modification: https://github.com/jonathanslenders/python-prompt-toolkit/commit/a454a26c5cd3ff7bd6d6d6a92f8b670908d84c32
If you have some time, can you check whether this solves your issue?

The reason for this change is that I want this fix to happen as well, even if ``select_first/select_last/insert_common_part`` are given.

randy3k:
Thanks. I haven't tested it but I believe that it should resolve my issue.

PR #612, opened 2018-04-21
need to refresh ui after resetting completion state

randy3k:
a continuation of #611

### DIFF
--- a/prompt_toolkit/buffer.py
+++ b/prompt_toolkit/buffer.py
@@ -1476,6 +1476,9 @@ class Buffer(object):
                 # When there are no completions, reset completion state anyway.
                 if not completions:
                     self.complete_state = None
+                    # Render the ui if the completion menu was shown
+                    # it is needed especially if there is one completion and it was deleted.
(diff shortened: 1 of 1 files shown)


jonathanslenders:
Hi @randy3k,

Can you tell me what goes wrong without this change? Why it is needed.
I would probably like to add one or two lines documentation.

Thanks!
Jonathan

randy3k:
If there was only one completion item and [this line](https://github.com/randy3k/rtichoke/blob/a3f204dad85212565ab50950b05817fd96bd099c/rtichoke/deps/prompt_toolkit/buffer.py#L1467-L1467) deleted it. The completion menu will displayed, but the `complete_state` is [set to](https://github.com/randy3k/rtichoke/blob/a3f204dad85212565ab50950b05817fd96bd099c/rtichoke/deps/prompt_toolkit/buffer.py#L1478-L1478) `None` . We end up with the menu poping up but the state is empty.

jonathanslenders:
Perfect! Thanks a lot @randy3k!

PR #615, opened 2018-04-26
Fixes for 2.0

asmeurer:
Some fixes for the 2.0 branch as I am trying to upgrade. 

Not sure how much I'll push here. Feel free to merge at any time.

### DIFF
--- a/prompt_toolkit/formatted_text/pygments.py
+++ b/prompt_toolkit/formatted_text/pygments.py
@@ -19,6 +19,9 @@ class PygmentsTokens(object):
         result = []
 
         for token, text in self.token_list:
-            result.append(('class:' + pygments_token_to_classname(token), text))
+            if list(token) == ['ZeroWidthEscape']:
+                result.append(('[ZeroWidthEscape]', text))
+            else:
+                result.append(('class:' + pygments_token_to_classname(token), text))
 
         return result
--- a/docs/pages/upgrading.rst
+++ b/docs/pages/upgrading.rst
@@ -152,7 +152,8 @@ adaptor to make it compatible with the native prompt_toolkit objects.
 - If you have a Pygments style, then this needs to be converted as well. A
   Pygments style class can be converted in a prompt_toolkit
   :class:`~prompt_toolkit.styles.Style` with the
-  :func:`~prompt_toolkit.styles.pygments.style_from_pygments` function. A
+  :func:`~prompt_toolkit.styles.pygments.style_from_pygments_cls` function
+        (which used to be called ``style_from_pygments``). A
   Pygments style dictionary can be converted using
   :func:`~prompt_toolkit.styles.pygments.style_from_pygments_dict`.
 
--- a/prompt_toolkit/output/defaults.py
+++ b/prompt_toolkit/output/defaults.py
@@ -18,8 +18,7 @@ def create_output(stdout=None):
     Return an :class:`~prompt_toolkit.output.Output` instance for the command
     line.
 
-    :param color_depth: Any `ColorDepth` value, a callable that returns a
-        `ColorDepth` or `None` for default.
+    :param stdout: The stdout object
     """
     stdout = stdout or sys.__stdout__
 
--- a/prompt_toolkit/shortcuts/prompt.py
+++ b/prompt_toolkit/shortcuts/prompt.py
@@ -327,8 +327,8 @@ class Prompt(object):
             return to_filter(value)()
         return dynamic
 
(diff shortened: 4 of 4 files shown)


asmeurer:
OK, I've refactored the Prompt class a bit. Take a look. This should make it much easier to customize the defaults by subclassing Prompt and redefining the methods for the parts I want to change, without having to copy everything that I want to keep the same. 

I haven't tested it yet (I am still getting my code to work), so please test it before merging. Also please read the comments in the commit message.

jonathanslenders:
Hi @asmeurer,

Thanks a lot for all of this!

``extra_input_processors `` was renamed to ``input_processors``. I dropped the "extra_" prefix for consistency and not to bother people with the internal implementation. (There are predefined key bindings, styles, etc... for none of them I used the "extra_" prefix.)

Just wondering: which customizations were you looking for that required overriding all these methods?

If it makes sense, I'll probably merge it, but +1 on keeping them private for now.

asmeurer:
Ah, I was confused about `input_processors`. I see now that they are indeed added to the default ones. 

> Just wondering: which customizations were you looking for that required overriding all these methods?

I use my own buffer subclass (see https://github.com/asmeurer/mypython/blob/master/prompt_toolkit_extensions.md). I also create my own application instance (I don't remember why; I haven't gotten far enough yet with 2.0 to see if it's still needed). I might at some point make my own layout as well. 

> If it makes sense, I'll probably merge it, but +1 on keeping them private for now.

I should probably figure out the test failures first.

asmeurer:
I fixed the tests. Some of the new attributes needed to be private (they were already being used). I'm not sure which of the attributes should be private or not. Should I rename things so that none of them are? Or should I make them all private?

asmeurer:
I don't feel very strongly either way about this, but it seems to me the `Prompt` class should really be called `Session`. It maintains state between multiple prompts, so that seems to be what it is (correct me if I am wrong).

prompt_toolkit/formatted_text/pygments.py
```diff
@@ -19,6 +19,9 @@ def __pt_formatted_text__(self):
         result = []
 
         for token, text in self.token_list:
-            result.append(('class:' + pygments_token_to_classname(token), text))
+            if list(token) == ['ZeroWidthEscape']:
```
jonathanslenders: Why is this needed? A Pygments lexer should never yield "ZeroWidthEscape".
The idea of this class is only to provide compatibility with whatever Pygments generates.

asmeurer:
That sounds fine. I might not get to it til next week.

asmeurer:
OK, I see you already did it. If you could decide either way on the remaining commits here that would be great. I've got code written that depends on the changes here so if any of them aren't going to be accepted that would be good to know so I can fix my code.

jonathanslenders:
@asmeurer: I merged most other changes in this commit: https://github.com/jonathanslenders/python-prompt-toolkit/commit/84902965e4fca5f9c837705fb9e400386e7523d3

One thing I'm not entirely sure I want to do is are the changes in ``PygmentsTokens``. If we do this, it probably needs to go in ``pygments_token_to_classname``, and probably we also have to memoize that function for performance as well.

asmeurer:
The zero width escape thing is there for backwards compatibility. Without it, it's not possible to mix it with pygments tokens (unless I'm mistaken). It would also break any end-user config that uses it. It's not really clear to me how it can go in pygments_token_to_classname, since that returns the part after `class:` and the zerowidthtoken is just `[ZeroWidthEscape]` without `class:`. It would definitely be simpler if it used the same format, that is, if `class:pygments.zerowidthescape` were recognized as a zero width escape token. Should that be the change instead? 

By the way, relating to memoizing, consider

```py
from prompt_toolkit.shortcuts.prompt import PromptSession

i = 0
def prompt_text():
    global i
    i += 1
    return str(i) + '> '

p = PromptSession(prompt_text)

while True:
    out = p.prompt()
    print(out)
```

Just pressing Enter here outputs

```
24>

48>

72>

96>

112>
```

And every time you type text, the prompt changes. What are the guarantees that prompt-toolkit makes about how often a prompt function will be called to update the prompt? I can imagine someone putting a calculation in a prompt function not realizing that it will be called dozens of times for each prompt.

asmeurer:
@jonathanslenders how do you feel about the zerowidthescape changes from here? Without them, they break backwards compatibility for anyone using a them in custom prompt. 

IIRC without the change one cannot use the zero width escape token with the pygments compatibility layer (but I could be wrong about that).

PR #619, opened 2018-05-10
Fixed selection for emacs mode. (Unlike Vi, Emacs doesn't include the…

jonathanslenders:
… upper boundary in the selection.)


asmeurer:
This looks like the correct behavior for emacs mode. I didn't test vi mode.

asmeurer:
Actually there is a bug here. The last character of each line is not visually selected (it is actually part of the visual selection, it just isn't shown visually).

prompt_toolkit/layout/processors.py
```diff
@@ -219,7 +219,7 @@ def apply_transformation(self, transformation_input):
                 # visualise the selection.
                 return Transformation([(selected_fragment, ' ')])
             else:
-                for i in range(from_, to + 1):
+                for i in range(from_, to):
```
asmeurer: I think this change should not be here. If I undo it, the last character is selected again. 

As I noted in the issue, the visual selection (in master) matches what is selected just fine.  The issue is just what gets selected, particularly when a selection is started.

PR #620, opened 2018-05-13
Add dictionary search in RadioList

gpotter2:
Hello,

This is my first contribution to prompt_toolkit, so it surely is a piece of crap. I've implemented it with what I knew.

This PR:
- adds a key binder so that when one type a letter while having a `RadioList` opened, it automatically switches to the next entry starting with that letter. This behavior is the same than in explorer for instance.

Demo code:

```python
from __future__ import unicode_literals
from prompt_toolkit.formatted_text import HTML
from prompt_toolkit.shortcuts.dialogs import radiolist_dialog
result = radiolist_dialog(
    values=[
        ('red', 'Red'),
        ('green', 'Green'),
        ('blue', 'Blue'),
        ('blue2', 'Blue2'),
        ('blue3', 'Blue'),
        ('bluec', 'blue. hi, im case sensitive'),
        ('orange', 'Orange'),
        ('orange', 'Orange2'),
        ('blue4', 'Blue again'),
    ],
    title='Radiolist dialog example',
    text='Please select a color:'
)
```

Then type, for instance, "B" (note the cap letter), several times

Surely, you all know a better way to do this, using what already exist. If you have any good advices on how to do this in a better way, please tell me

### DIFF
--- a/prompt_toolkit/widgets/base.py
+++ b/prompt_toolkit/widgets/base.py
@@ -31,6 +31,7 @@ from prompt_toolkit.layout.margins import ScrollbarMargin, NumberedMargin
 from prompt_toolkit.layout.processors import PasswordProcessor, ConditionalProcessor, BeforeInput
 from prompt_toolkit.mouse_events import MouseEventType
 from prompt_toolkit.utils import get_cwidth
+from prompt_toolkit.keys import Keys
 
 from .toolbars import SearchToolbar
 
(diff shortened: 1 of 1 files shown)


jonathanslenders:
Hey @gpotter2,

This is a cool idea. I like it!
I had some remarks with a couple of things that can be improved before merging.

I'm not sure I'll be able to merge it before the next release, but it's definitely a valuable contribution.

gpotter2:
Thanks !

This can definatly wait :)

I'm really awaiting for the next release ! It will unblock IPython from supporting it, and will greatly help us. We're stuck between choosing `prompt_toolkit>=2.0.0` or `IPython`

PR #621, opened 2018-05-14
Added ThreadedHistory for loading the history asynchronously.

jonathanslenders:


prompt_toolkit/history.py
```diff
@@ -127,39 +223,3 @@ def write(t):
...
-        return self.strings[key]
-
-    def __iter__(self):
-        return iter(self.strings)
-
-    def __len__(self):
-        return len(self.strings)
-
-
-class DynamicHistory(History):
```
asmeurer: I've been using DynamicHistory to hot swap in a DummyHistory when I don't want an item to be saved (for https://github.com/jonathanslenders/python-prompt-toolkit/issues/610 in particular). How would you recommend doing this with DynamicHistory removed?

jonathanslenders: Yes, I've been thinking about this, but swapping of history backends in the middle of a running prompt() function doesn't work, due to how the load callbacks from the history are now attached to the Buffer.

Why do you want to hot swap it? Would it be fine to hot swap it in between two prompt() calls?

jonathanslenders: The work-around is to create a new ``PromptSession`` if you need a different history object.
I haven't seen any application swapping the history backend at runtime, so I'm curious why this is needed.

jonathanslenders: Right now, the easiest would be indeed to wrap the history class, override ``append_string`` and ignore anything that doesn't need to be stored.

PR #627, opened 2018-06-02
Add dictionary search in RadioList

gpotter2:
### This PR is a rebase of https://github.com/jonathanslenders/python-prompt-toolkit/pull/620

Hello,

This is my first contribution to prompt_toolkit. I've implemented it with what I knew.

This PR:
- adds a key binder so that when one type a letter while having a `RadioList` opened, it automatically switches to the next entry starting with that letter. This behavior is the same than in explorer for instance.

Demo code:

```python
from __future__ import unicode_literals
from prompt_toolkit.formatted_text import HTML
from prompt_toolkit.shortcuts.dialogs import radiolist_dialog
result = radiolist_dialog(
    values=[
        ('red', 'Red'),
        ('green', 'Green'),
        ('blue', 'Blue'),
        ('blue2', 'Blue2'),
        ('blue3', 'Blue'),
        ('bluec', 'blue. hi, im case sensitive'),
        ('orange', 'Orange'),
        ('orange', 'Orange2'),
        ('blue4', 'Blue again'),
    ],
    title='Radiolist dialog example',
    text='Please select a color:'
)
```

Then type, for instance, "B" (note the cap letter), several times

 If you have any good advices on how to do this in a better way, please tell me

### DIFF
--- a/prompt_toolkit/widgets/base.py
+++ b/prompt_toolkit/widgets/base.py
@@ -31,6 +31,7 @@ from prompt_toolkit.layout.margins import ScrollbarMargin, NumberedMargin
 from prompt_toolkit.layout.processors import PasswordProcessor, ConditionalProcessor, BeforeInput
 from prompt_toolkit.mouse_events import MouseEventType
 from prompt_toolkit.utils import get_cwidth
+from prompt_toolkit.keys import Keys
 
 from .toolbars import SearchToolbar
 
@@ -507,6 +508,14 @@ class RadioList(object):
         @kb.add(' ')
         def _(event):
             self.current_value = self.values[self._selected_index][0]
+
+        @kb.add(Keys.Any)
+        def _(event):
+            # We first check values after the selected value, then all values.
+            for value in self.values[self._selected_index + 1:] + self.values:
+                if value[1].startswith(event.data):
+                    self._selected_index = self.values.index(value)
+                    return
 
         # Control and window.
         self.control = FormattedTextControl(


prompt_toolkit/widgets/base.py
```diff
@@ -508,6 +509,22 @@ def _(event):
         def _(event):
             self.current_value = self.values[self._selected_index][0]
 
+        @kb.add(Keys.Any)
+        def _(event):
+            if six.binary_type(self.values[self._selected_index][1]).startswith(event.data):
```
jonathanslenders: Hi @gpotter2,
Did you test this on Python 3? I think ``event.data`` is a ``str``, not ``bytes``. Normally this conversion is not needed.

prompt_toolkit/widgets/base.py
```diff
@@ -508,6 +509,22 @@ def _(event):
...
+                # In the case we are already on a value starting by the key: seek for the next one
+                try:
+                    self._selected_index = self.values.index(next(x for x in self.values[self._selected_index + 1:]
```
jonathanslenders: For readability, I'd prefer a normal for-loop with a break, instead of using ``next``. Something like:
```python
for value in self.values[self._selected_index + 1:] + self.values:
    if value.startswith(event.data):
        self._selected_index = self.values.index(value)
        break
```
Adding ``self.values`` here helps searching again from the beginning.

jonathanslenders:
Hi @gpotter2, I added two remarks. If you have some time to look at that, I can merge it. Thanks again!

prompt_toolkit/widgets/base.py
gpotter2: Hi! You were right, that was not needed, I thought remembering having troubles with unicode being retrieved on Python 2, but it's working as so. Just updated the code

prompt_toolkit/widgets/base.py
gpotter2: Good idea for the `+ self.values` ! Updated

gpotter2:
Just updated the PR according to your comments

prompt_toolkit/widgets/base.py
```diff
@@ -508,6 +509,15 @@ def _(event):
         def _(event):
             self.current_value = self.values[self._selected_index][0]
 
+        @kb.add(Keys.Any)
+        def _(event):
+            # We first check values after the selected value, then all values
+            for value in self.values[self._selected_index + 1:] + self.values:
+                if value[1].startswith(event.data):
+                    self._selected_index = self.values.index(value)
+                    return True
```
jonathanslenders: One last remark: key bindings return None. So a simple return ``return`` here would do it, and you can remove the next line. Thanks!

prompt_toolkit/widgets/base.py
```diff
@@ -508,6 +509,15 @@ def _(event):
         def _(event):
             self.current_value = self.values[self._selected_index][0]
 
+        @kb.add(Keys.Any)
+        def _(event):
+            # We first check values after the selected value, then all values
```
jonathanslenders: (And if you find the time to fix the comment below. Usually I terminate sentences with a dot. ;) It wouldn't keep me from merging it though.)

gpotter2:
Updated

jonathanslenders:
Thanks a lot, @gpotter2!

PR #631, opened 2018-06-05
Fix Opt+left/right in iterm

### LINKED ISSUE #483: iterm option+left/right not handled
In iTerm2 (3.0.14) I run ipython (5.2.0, using PT) and get this when I option+left/right:

```
In [1]: awoifej woeifj wefoiwj ef[1;9D[1;9D[1;9C[1;9C
```

As you can see, option+left sends `[1;9D` and option+right sends `[1;9C`.

In the same ipython in OSX Terminal, these work correctly and map to word-left and word-right. Possibly OSX Terminal sends different escape codes. 

I found this comment in the source code: https://github.com/jonathanslenders/python-prompt-toolkit/blob/master/prompt_toolkit/terminal/vt100_input.py#L178-L190

which says those keys are handled in the input processor, but I was unable to find any such handling.

I don't believe I have any overrides for those keys in my iterm2 profile (I've reset to 'xterm defaults'). In Profiles -> Keys I have both Option keys set to 'Normal': 
screen shot 2017-03-07 at 10 41 15 am

If I instead load the preset 'Terminal.app Compatability' it works around the issue. I note that this adds these two hex code overrides for the keys:
screen shot 2017-03-07 at 10 44 47 am

craigds:
Fixes #483 using the fix @jonathanslenders suggested.

I actually tested this in 1.0.15 by putting the fix in `prompt_toolkit/terminal/vt100_input.py` instead of here.

I wasn't able to manually test this with prompt_toolkit 2.0 since I haven't found a version of ipython that will use it. However I'd assume it doesn't change materially in 2.0


jonathanslenders:
This looks good. According to the following comment, there are some other sequences that we would have to support: https://github.com/fish-shell/fish-shell/issues/3056#issuecomment-221035019
Maybe I should have a look at the Fish source.

jonathanslenders:
See: https://github.com/fish-shell/fish-shell/blob/eca4273f7777ba01a59a40c2e45f4f04561421ce/share/functions/__fish_shared_key_bindings.fish#L60

PR #638, opened 2018-06-13
Fix color names in printouts to match tags

cppguru:


jonathanslenders:
Thanks again!

cppguru:
You are very welcome.

PR #650, opened 2018-06-26
Added buffer argument to accept_handler of TextArea

Arrowbox:
This should allow for the accept handler to inspect
the buffer from the TextArea.


jonathanslenders:
Thank you @Arrowbox, this is great!

PR #651, opened 2018-06-26
Added `default` parameter to RadioList and radiolist_dialog

Arrowbox:
Default index allows for pre-selecting an entry in a list,
useful for dialogs with many values.

I've found this useful if the dialog is updating something that was already selected previously.

### DIFF
--- a/prompt_toolkit/shortcuts/dialogs.py
+++ b/prompt_toolkit/shortcuts/dialogs.py
@@ -111,14 +111,14 @@ def message_dialog(title='', text='', ok_text='Ok', style=None, async_=False):
 
 
 def radiolist_dialog(title='', text='', ok_text='Ok', cancel_text='Cancel',
-                     values=None, style=None, async_=False):
+                     values=None, style=None, async_=False, default=0):
     """
     Display a simple message box and wait until the user presses enter.
     """
     def ok_handler():
         get_app().exit(result=radio_list.current_value)
 
-    radio_list = RadioList(values)
+    radio_list = RadioList(values, default)
 
     dialog = Dialog(
         title=title,
--- a/prompt_toolkit/widgets/base.py
+++ b/prompt_toolkit/widgets/base.py
@@ -497,16 +497,19 @@ class RadioList(object):
     List of radio buttons. Only one can be checked at the same time.
 
     :param values: List of (value, label) tuples.
+    :param default: Default index to select, defaults to 0
     """
-    def __init__(self, values):
+    def __init__(self, values, default=0):
         assert isinstance(values, list)
         assert len(values) > 0
         assert all(isinstance(i, tuple) and len(i) == 2
                    for i in values)
+        assert 0 <= default < len(values)
 
         self.values = values
-        self.current_value = values[0][0]
-        self._selected_index = 0
+
+        self.current_value = values[default][0]
+        self._selected_index = default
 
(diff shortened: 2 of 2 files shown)


Arrowbox:
@jonathanslenders Do I need to open an issue for you to consider this addition?

tests/test_radiolist.py
```diff
@@ -0,0 +1,96 @@
...
+                (1, 'some_text'),
+                ('fizz', 'some_more_text'),
+                ({'foo':'bar'}, 'even_more_text')
+             ]
+    radiolist = RadioList(values, 1)
+    assert radiolist.current_value == 'fizz'
+    assert radiolist._selected_index == 1
+
+def test_bad_params():
+    with pytest.raises(AssertionError):
```
jonathanslenders: Hi @Arrowbox,
Thanks a lot for writing unit tests, but for these, we shouldn't ever assume that ``AssertionError`` is ever raised. If Python runs with the ``-O`` option, then no assertions are ever raised. Removing an assertion from the code is not supposed to break something, so there can't be a unit test which states it does.

I think raising ``ValueError`` is more appropriate, if we want to have some flow control in case of passing a bad input.

prompt_toolkit/shortcuts/dialogs.py
```diff
@@ -111,14 +111,14 @@ def message_dialog(title='', text='', ok_text='Ok', style=None, async_=False):
 
 
 def radiolist_dialog(title='', text='', ok_text='Ok', cancel_text='Cancel',
-                     values=None, style=None, async_=False):
+                     values=None, style=None, async_=False, default=0):
```
jonathanslenders: I like the ``default`` argument, but I would prefer if it was not an integer, but a key from the ``values`` list. Would it be possible to change it that way. The default can be ``None`` which always takes the first value.

tests/test_radiolist.py
Arrowbox: I'm happy to change that, but there are a bunch of other `assert precondition` statements at the beginning of the function already. I was trying to follow the previous format and thought an extra test there would be okay. Should I change all the checks? Or maybe just drop the test?

prompt_toolkit/shortcuts/dialogs.py
Arrowbox: I'd hesitate because the input is a list of tuples. There is no guarantee of uniqueness on the values or key. Maybe also accept the tuple from the list? And do a check internally and decide what to do?

```python
if default is None:
  self.current_value = 0
elif isinstance(default, tuple):
  self.current_value = values.index(default)
else:
  self.current_value = default
```

tests/test_radiolist.py
Arrowbox: I looked at testing if assert has been stripped to decide if we should just drop the test but that did not seem trivial.

gpotter2:
Hi ! Cool PR. What’s its state ? I’m also looking forward to this.

I’ve also been reworking RadioList in https://github.com/jonathanslenders/python-prompt-toolkit/pull/758 so I’m expecting some conflicts.

jonathanslenders:
@gpotter2: I believe I had some small reservations regarding these changes, but didn't had the time yet to really think about it. In any case, I think it's useful and it'll be merged in one way or the other.
I hope to get some time this weekend to work on it.

PR #669, opened 2018-07-10
CI: use setup.cfg, -U and inline commands

lwm:
Reads better to me but perhaps there was a reason everything was broken out onto new lines?


jonathanslenders:
Merged. Thank you @lwm !

PR #683, opened 2018-07-13
Making more pygments lexers imports explicit

tsvenson:
I think I have found all places in code and doc where importing of pygments lexers needed to be more explicit to avoid making IDE's like PyCharm complain.


lwm:
> more explicit to avoid making IDE's like PyCharm complain.

Do you know what is making those errors (flake8 or something of the like)? We could add to the CI.

jonathanslenders:
Thanks!

PR #688, opened 2018-07-18
Don't start a timeout thread if already waiting

### LINKED ISSUE #687: Large number of threads created when run under PyCharm
request_absolute_cursor_position keeps getting called when running a simple app which starts a new thread if the CPR state is unknown (default) to do a timeout set of NOT_SUPPORTED.  However, there's no check to enforce that only 1 such timeout thread exists.  Running in PyCharm with a debugger can easily get >100 threads created & completely bog down the debugger.

vlovich:
Only start 1 timeout thread ever to set state to NOT_SUPPORTED.  Avoids using extra resources unnecessarily & avoids cpr_not_supported_callback being called multiple times if it isn't supported (since there's no lock being held when self.cpr_support is being checked/modified).

Fixes #687

### DIFF
--- a/prompt_toolkit/renderer.py
+++ b/prompt_toolkit/renderer.py
@@ -388,11 +388,12 @@ class Renderer(object):
                 return
             else:
                 # Asks for a cursor position report (CPR).
+                start_timer = len(self._waiting_for_cpr_futures) == 0
                 self._waiting_for_cpr_futures.append(Future())
                 self.output.ask_for_cpr()
 
                 # If we don't know whether CPR is supported, test using timer.
-                if self.cpr_support == CPR_Support.UNKNOWN:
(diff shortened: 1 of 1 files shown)


jonathanslenders:
This looks good. Thank you @vlovich! I'll merge it asap. (Normally this weekend.)

jonathanslenders:
Hi @vlovich,

I think that in this case there is more to do. We should not start the timer, but also not call `output.ask_for_cpr`. I made a different merge request: https://github.com/jonathanslenders/python-prompt-toolkit/pull/691 Do you think that's fine. Could you maybe give that a try as well?
Thanks a lot!

jonathanslenders:
Fixed by: https://github.com/jonathanslenders/python-prompt-toolkit/pull/691

Issue #168, opened 2015-09-29
Mouse support breaks scrolling


mdraw:
As I mentioned here https://github.com/scopatz/xonsh/pull/393#issuecomment-143664623, enabling the new mouse support makes scroll wheel scrolling impossible.

Code to reproduce:

    from prompt_toolkit.shortcuts import get_input
    get_input(mouse_support=True)

Execute and try using your scroll wheel in the terminal. Does it work on your system?

jonathanslenders:
Hi @mdraw,

It depends. By default, it's impossible to to both capture mouse events and support scrolling. (That would be ambiguous.) However, it's up to the terminal application to decide whether or not forward the scrolling event to the pseudo terminal. E.g. in tmux it's always possible (Press ``Ctrl-B`` followed by ``[``.) In some terminals, scrolling still works when shift or option is pressed. (E.g. shift for urxvt.)

We output a vt100 instruction that requests the terminal to send button clicks to stdin. Unfortunately scrolling is considered a click as well, (using button 4 and 5.)

So, I guess not much to do about this. Better to make it configurable and allow the user to decide whether or not to handle mouse events.

Jonathan

mdraw:
Thanks for the explanation! Good to know that tmux scrolling with `[` still works, I hadn't tried that.

Looks like this is a trade-off and a matter of personal preference, so I completely agree with you that it's a nice feature as long as it is optional and the potential problems with it are well-documented and easily visible by users.

Issue #686, opened 2018-07-18
Mouse scroll fails when wrap_lines=True for Window


Arrowbox:
I've not completely nailed down what the exact issue is. But when the buffer has many lines that are longer than the width and wrap then the mouse scroll sometimes gets stuck. Using keyboard to move the cursor beyond the edges does work but mouse scroll just doesn't respond.

Arrowbox:
Okay, I've narrowed it down 

```python
@property
def content_height(self):
        """
        The full height of the user control.
        """
        return self.ui_content.line_count
```

When line wrap is on, the `ui_content.line_count` does not match the number of lines available to display. So in the scroll down function:

```python
    def _scroll_down(self):
        " Scroll window down. "
        info = self.render_info

        if self.vertical_scroll < info.content_height - info.window_height:
            if info.cursor_position.y <= info.configured_scroll_offsets.top:
                self.content.move_cursor_down()

            self.vertical_scroll += 1
```

the `self.vertical_scroll` reaches the limit of the comparison before the actual end. I tried calculating the actual available height when the `RenderInfo` is initialized but that caused too much overhead.

I also removed the check before changing the scroll. The `do_scroll` function does adjust if we are out of bounds. It doesn't look like the performance takes a hit.

PR #693, opened 2018-07-25
Fix line wrap mouse scroll

Arrowbox:
For issue #168  
This fixes some of the mouse scrolling issues with Window. And adds some unit tests to check them.

* Can't scroll to end with `wrap_lines`
* Can't scroll past end with `allow_scroll_beyond_bottom`

Still needs work to get scrolling up to work. It currently stops scrolling up when the amount of lines needed to add a line from the top would force the cursor off screen after moving up 1 line. Or if the cursor is not on the last line of the window but on the last line of the visible lines.

### DIFF
--- a/prompt_toolkit/layout/containers.py
+++ b/prompt_toolkit/layout/containers.py
@@ -2014,10 +2014,17 @@ class Window(Container):
         info = self.render_info
 
         if self.vertical_scroll < info.content_height - info.window_height:
+            should_scroll = True
+        elif self.allow_scroll_beyond_bottom() and self.vertical_scroll < info.content_height:
+            should_scroll = True
+        else:
+            should_scroll = False
+
+        if should_scroll:
+            self.vertical_scroll += 1
             if info.cursor_position.y <= info.configured_scroll_offsets.top:
                 self.content.move_cursor_down()
 
-            self.vertical_scroll += 1
 
     def _scroll_up(self):
         " Scroll window up. "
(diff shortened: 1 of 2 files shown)


Arrowbox:
My mistake, this is actually for issue #686

Arrowbox:
@jonathanslenders It looks like there is some error with Appveyor and Python 2.6. It looks like some sort of SSL error with respect to pypi. My rudimentary googling seems to indicate an issue with python 2.6 and pip. Any thoughts? I don't think my changes would cause it.

jonathanslenders:
Don't worry about Python 2.6, especially on Windows there's no reason to use it (IMHO). It shouldn't be because of these changes.

Thanks for the pull request, I hope to get some time soon to review and merge it!

Arrowbox:
I've been working on the `_scroll_up` implementation and can't seem to find a work around for one particular issue. The problem is that the `Window._mouse_handler` can be called multiple times between calls to`Window.write_to_screen`. The `_mouse_handler` uses `Window.render_info` that is generated for each rendering pass. Included in the `render_info` is the `cursor_position` and `ui_content.cursor_position`.

Both `_scroll_down` and `_scroll_up` use the positions to determine if a call to `self.content.move_cursor_up|down` should be made. However, any call to those are not reflected in `self.render_info` until after the next render. Multiple calls to the one of the scroll functions only end up with a modified `self.vertical_scroll`.

I think a possible long term path would be to make the mouse scrolling functions set a `self.window_scroll` flag to `True`. The `Window._scroll` function could then check the flag to decide if the scroll offsets should be adjusted to fit the cursor position or if the cursor position should be adjusted to fit the scroll window. What do you think?

jonathanslenders:
Great analysis! Thanks a lot. I haven't had the time yet to dive into this issue, I will. It's something I think needs to be fixed.

PR #694, opened 2018-07-25
Updates path to Completer in autocompletion.py example.

6r1d:
Autocompletion example wasn't working, so I changed Completer according to current documentation. Hope it helps!


jonathanslenders:
Thank you @6r1d, I really appreciate it!

PR #706, opened 2018-08-10
add freud to full screen applications

stloma:
Hi there, thought I'd add my application to list that has benefited from your library!

Thanks again for your work!


jonathanslenders:
That looks really nice!

PR #712, opened 2018-08-24
progress_bar.formatters.Bar needs to compute length of start, end, and sym_b

njalerikson:
No assumptions are made about how long the start and end patterns should be. As a consequence we need to then also format the resulting Bar based on how long the start and end patterns are, not hardcode their widths to 1.

### DIFF
--- a/prompt_toolkit/shortcuts/progress_bar/formatters.py
+++ b/prompt_toolkit/shortcuts/progress_bar/formatters.py
@@ -129,8 +129,10 @@ class Bar(Formatter):
         self.sym_c = sym_c
         self.unknown = unknown
 
+        self._reserved_width = len(start) + len(sym_b) + len(end)
+
     def format(self, progress_bar, progress, width):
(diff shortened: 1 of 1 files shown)


jonathanslenders:
Great catch! But almost correct. It should use wcwidth, otherwise it doesn't work for double width (chinese) characters.

```python
from prompt_toolkit.utils import get_cwidth

# Now use get_cwidth instead of len.
```

Can you update the pull request and see whether it still works? Then I'll merge it.

jonathanslenders:
Hi @njalerikson,

Thanks for fixing the commit! I had to do two small modification:
- Storing the reserved width in ``self`` is not such a good idea, because it doesn't update if somebody assigns ``start/sym_b/end``.
- The change assumed that ``sym_b`` can have a length > 1, so I removed that assertion.

I edited and push the commit here: https://github.com/jonathanslenders/python-prompt-toolkit/commit/85fcb327c997f0bf3331d3470e7835e84d79486c

PR #717, opened 2018-09-04
Added get_line_prefix attribute to Window.

jonathanslenders:
This is a refactoring that opens up many possibilities.
- Line wrapping (soft or hard) can insert or remove whitespace in front of the
  line, or insert some symbols in front. Like the Vim "breakindentopt" option.
- Single line prompts also support line continuations now.
- Line continuations can have a variable width.

### DIFF
--- a/examples/full-screen/simple-demos/line-prefixes.py
+++ b/examples/full-screen/simple-demos/line-prefixes.py
@@ -1,0 +1,100 @@
+#!/usr/bin/env python
+"""
+An example of a BufferControl in a full screen layout that offers auto
+completion.
+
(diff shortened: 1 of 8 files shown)


jonathanslenders:
I'm glad you got it fixed! Sorry about breaking your code! I considered this an internal API indeed. https://xkcd.com/1172/

takluyver:
Thanks, I'm all too familiar with the scenario of that XKCD. I've opened an issue to figure out a better way to use prompt_toolkit: https://gitlab.com/takluyver/enboard/issues/1

PR #720, opened 2018-09-07
Don't restore editing_mode

meeuw:
editing_mode is saved in Application which isn't a property of
PromptSession and shouldn't be restored.

This prevents that editing_mode is restored for the next prompt in
a PromptSession for pgcli/mycli where editing_mode is changed
by a key binding while PromptSession.prompt is being executed.

### DIFF
--- a/prompt_toolkit/shortcuts/prompt.py
+++ b/prompt_toolkit/shortcuts/prompt.py
@@ -716,6 +716,7 @@ class PromptSession(object):
         def restore():
             " Restore original settings. "
             for name in self._fields:
+                if name == "editing_mode": continue
                 setattr(self, name, backup[name])
 
         def pre_run():


jonathanslenders:
Thank you @meeuw! Let me think about this for a little bit to see whether this is the right fix. Probably I'll merge it.

jonathanslenders:
Hi @meeuw,

As said, I'm hesitating a bit here. The reason is consistency with other attributes of ``PromptSession``.

The problem is as follows. ``PromptSession`` is a wrapper that glues everything for a prompt application together. This means, it's a higher level abstraction that contains the ``Application`` object. However, a user of ``PromptSession`` builds a prompt, by assigning attributes to ``PromptSession`` directly. Many of the underlying ``Application`` attributes refer to the attributes of this ``PromptSession``, (using ``DynamicCompleter``, ``DynamicLexer``, ...), so that every modification of ``PromptSession`` is then reflected in the ``Application`` object.
For some attributes however, this is not possible: ``editing_mode``, ``vi_mode``, ``color_depth``, ``tempfile_suffix``, etc... because they are strings or enumerations. (There is some degree of reactive programming here, but as you see, it doesn't work everywhere or in any direction.)

The second problem is that arguments can be passed to both the ``PromptSession`` ``__init__``, as well as the ``prompt()`` method. Whatever the method receives will override the ``__init__`` attributes, but only for that call.

So, handling ``editing_mode`` as a special case is - I think - not the right approach. However, I think the following could maybe work. I made another pull request: https://github.com/jonathanslenders/python-prompt-toolkit/pull/732

What do you think? Can you check whether this also resolves your issue?
Thanks!

meeuw:
#732 also fixes the issue for pgcli. To be honest, I'm not sure what the purpose is of the backup. Is it to allow one-time changes to the prompt session for `PromptSession.prompt()`? When is this useful?

jonathanslenders:
Actually, that's a fair comment. I think I'll remove the backup/restore functionality completely. I don't know of anyone using it. In fact,``PromptSession`` is already something stateful to begin with.

jonathanslenders:
The restore functionality has been removed: https://github.com/jonathanslenders/python-prompt-toolkit/commit/928379c56955c659ebd51b6e1a6b4a739a4ae241
If you have time, can you try again, and close this issue if it works? I plan a new release soon.

meeuw:
"Perfection is achieved, not when there is nothing more to add, but when there is nothing left to take away."

Antoine de Saint-Exupery
French writer (1900 - 1944)

Issue #305, opened 2016-05-29
Data grid / table


tony:
With a fix set of columns / dynamic number of rows, is there any proof of concepts out there for handling a tabular data?

jonathanslenders:
No, sorry. Right now, there is no proof of concept of this... I keep this issue open as a reminder for when I'm going to create more examples.

therden:
@tony:   Are you inquiring only about _displaying_ tabular data -- or about displaying an _editable_ data table -- using `prompt_toolkit`?  (Just curious...)

tony:
displaying

swapniljariwala:
Even I want something like this for one of my side project. I'll share the code if I complete this piece.

jonathanslenders:
@swapniljariwala, if you want to try. Start from the prompt_toolkit 2.0 branch. (that should be easier and will be released soon.) I think you have to implement a custom FormattedTextControl with custom key bindings. The scrolling can be done by wrapping it in a Window.

lwm:
Would https://github.com/nirum/tableprint cover your use case?

zokrezyl:
you can use https://rich.readthedocs.io/en/stable/tables.html
1. create rich.Console with StringIO output buffer
2. print the tabel to the console
3. get the vale of the buffer (buffer.get_value())
4. render it to terminal with an ANSI object

PR #724, opened 2018-09-13
Table Container

njalerikson:
I am aware that this PR is not finished yet. I want to get community feedback to see if there are any preferred ways of implementing this.

The implementation thus far works in generating a table. Still need to figure out the ideal way of controlling column widths.

A demo of the container is available as part of this PR:

```python
from prompt_toolkit.widgets import table

table.demo()
```

I was initially thinking to use the width logic used by `prompt_toolkit.shortcuts.progressbar.ProgressBar` but I don't think that will work given the desire to support merged cells.

#305

### DIFF
--- a/prompt_toolkit/widgets/table.py
+++ b/prompt_toolkit/widgets/table.py
@@ -1,0 +1,310 @@
+from prompt_toolkit import Application
+from prompt_toolkit.layout.layout import Layout
+from prompt_toolkit.widgets import Box, TextArea
+# from prompt_toolkit.widgets.base import Border
+from prompt_toolkit.layout.containers import Window, VSplit, HSplit, HorizontalAlign, VerticalAlign
+from prompt_toolkit.layout.dimension import Dimension
+from prompt_toolkit.key_binding.key_bindings import KeyBindings
+
+
+class SpaceBorder:
+    " Box drawing characters. (Spaces) "
+    HORIZONTAL = ' '
+    VERTICAL = ' '
+
+    TOP_LEFT = ' '
+    TOP_RIGHT = ' '
+    BOTTOM_LEFT = ' '
+    BOTTOM_RIGHT = ' '
+
+    LEFT_T = ' '
+    RIGHT_T = ' '
+    TOP_T = ' '
+    BOTTOM_T = ' '
+
+    INTERSECT = ' '
+
+
+class AsciiBorder:
+    " Box drawing characters. (ASCII) "
+    HORIZONTAL = '-'
+    VERTICAL = '|'
+
+    TOP_LEFT = '+'
+    TOP_RIGHT = '+'
+    BOTTOM_LEFT = '+'
+    BOTTOM_RIGHT = '+'
+
+    LEFT_T = '+'
+    RIGHT_T = '+'
+    TOP_T = '+'
+    BOTTOM_T = '+'
+
+    INTERSECT = '+'
+
+
+class ThinBorder:
+    " Box drawing characters. (Thin) "
+    HORIZONTAL = '\u2500'
+    VERTICAL = '\u2502'
+
+    TOP_LEFT = '\u250c'
+    TOP_RIGHT = '\u2510'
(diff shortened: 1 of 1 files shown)


njalerikson:
Here's an example of what this table container is capable of rendering.

```
 ╔═══════════════╦═══════════════╦═══════════════╦═══════════════╗
 ║Hello World    ║Praesent eu ult║Lorem ipsum dol║               ║
 ║               ║               ║or sit amet, co║               ║
 ║               ║               ║nsectetur adipi║               ║
 ║               ║               ║scing elit. Ut ║               ║
 ╠═══════════════╩═══════════════╬═══════════════╬═══════════════╣
 ║Buzz                           ║Morbi viverra, ║               ║
 ║                               ║justo eget pret║               ║
 ║                               ║ium sollicitudi║               ║
 ║                               ║n, magna ex sod║               ║
 ╠═══════════════╦═══════════════╩═══════════════╩═══════════════╣
 ║<The quick b   ║Ut egestas vel nisi et sodales. Etiam arcu mass║
 ║               ║a, viverra in pellentesque quis, molestie a lac║
 ║               ║us. Nulla suscipit mi luctus blandit dignissim.║
 ║               ║ Proin ac turpis sit amet enim luctus venenatis║
 ║               ║ quis et orci. Donec sed tortor ex. Class apten║
 ╠═══════════════╬═══════════════╦═══════════════╦═══════════════╣
 ║<Hello World   ║Donec placerat ║               ║               ║
 ║               ║lacus egestas, ║               ║               ║
 ║               ║aliquam enim vi║               ║               ║
 ║               ║tae, congue ips║               ║               ║
 ║               ║um. Praesent vi║               ║               ║
 ╠═══════════════╬═══════════════╬═══════════════╬═══════════════╣
 ║Buzz           ║Aliquam eleifen║               ║               ║
 ║               ║d mi arcu, sit ║               ║               ║
 ║               ║amet convallis ║               ║               ║
 ║               ║tellus condimen║               ║               ║
 ╚═══════════════╩═══════════════╩═══════════════╩═══════════════╝
```

jonathanslenders:
Hi @njalerikson,

This looks really nice!

Some people have been asking for this. (I think the pgcli developers would like something like this for rendering the query output. On top of that, one of the developers wanted a way to select a cell, edit the content, and save it again. Further, this would require some kind of scrolling. I guess that can be implemented on top of this.)

I still need to go through all the changes, but I will have some feedback. Before merging it, we have two options. Either we give this pull request some time for it to become stable and mature, or we release it as a separate package until the API is stable, and can be merged. (I don't like to introduce too many backwards-incompatible changes in prompt_toolkit, but at the same time I want code to be really nice before merging it.)

Anyway, thanks a lot for starting this!

Jonathan

njalerikson:
So I am currently pretty pleased with how the widths are handled. It works wonderfully when using weighted widths and/or fixed widths.

What I currently dislike is that there isn't a way to handle a "fitted" width where a column width is governed exclusively by the preferred width value and not expanded to fill the screen (as currently happens). How do you think that could be best accomplished? I would expect a table where all of the columns are being "fitted" could potentially result in a table that doesn't fill the screen width.

Finally, I think that there should be a different way of solving the need to override the `_divide_widths` method from `_Row`/`VSplit`. I dislike how it's almost a perfect copy of the original.

Maybe it makes more sense to modify `VSplit` to where we can pass in `column_width` and `column_widths` arguments there instead?

gpotter2:
If this is still on the table, now (prompt_toolkit 3.0 first commits) is definitely the time to

> introduce too many backwards-incompatible changes in prompt_toolkit at the same time

This feature would be great to have :smile:

Vosjedev:
are there still people working on this?

lyz-code:
Hi @Vosjedev in the past I started working on a different approach for the same solution. I started creating [`prompt-toolkit-table`](https://github.com/lyz-code/prompt-toolkit-table/tree/feat/initial_iteration) but never finished the first version. 

If anyone wants to continue where I left it I can transfer the repo or accept PR

PR #732, opened 2018-09-21
Only restore PromptSession attributes that were given to the prompt()…

jonathanslenders:
… method.


jonathanslenders:
See: https://github.com/jonathanslenders/python-prompt-toolkit/commit/928379c56955c659ebd51b6e1a6b4a739a4ae241

PR #738, opened 2018-09-25
Update README.rst

jameswinegar:
Include Python 3.6 in readme.

Assuming true based on tests that have been running since last year.


jonathanslenders:
Actually, it runs fine too on 3.7. Can you change it to 3.7?

PR #748, opened 2018-10-09
Add license to the setup.py

merqurio:
In order to make automatic checking of license compliance faster, add license to the setup arguments.


jonathanslenders:
Thanks!

PR #771, opened 2018-11-01
Update asking_for_input.rst

urain39:
fix typo.
> To persist a history to disk, use a `FileHistory` **instead instead** of the default `InMemoryHistory`. This history object can be passed either to a `PromptSession` or to the `prompt()` function.


jonathanslenders:
Thanks!!

PR #779, opened 2018-11-03
add LazyHistory.

urain39:
It's only saving the history when exiting, not each line(make it writing less).


urain39:
Looks like multiline has bug, reopen later.

PR #780, opened 2018-11-04
Add LazyHistory

urain39:
Add Lazy History.
Fix multiline input. https://github.com/prompt-toolkit/python-prompt-toolkit/pull/779

### DIFF
--- a/examples/prompts/history/lazy-history.py
+++ b/examples/prompts/history/lazy-history.py
@@ -1,0 +1,26 @@
+#!/usr/bin/env python
+"""
+Simple example of a CLI that keeps a persistent history of all the entered
+strings in a file. When you run this script for a second time, pressing
+arrow-up will go back in history.
+"""
+from __future__ import unicode_literals
+from prompt_toolkit import PromptSession
+from prompt_toolkit.history import LazyHistory
+
+
+def main():
+    # The history needs to be passed to the `PromptSession`. It can't be passed
+    # to the `prompt` call because only one history can be used during a
+    # session.
+
+    with LazyHistory('.example-history-file') as our_history:
+        session = PromptSession(history=our_history)
+
+        while True:
+            text = session.prompt('Say something: ')
+            print('You said: %s' % text)
+
+
+if __name__ == '__main__':
+    main()
--- a/prompt_toolkit/history.py
+++ b/prompt_toolkit/history.py
@@ -252,7 +252,7 @@ class LazyHistory(FileHistory):
             return  # Skip empty
 
         with open(self.filename, 'wb') as f:
-            lines = ['\n# {0}\n'.format(datetime.now())]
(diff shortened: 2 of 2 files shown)


urain39:
Sorry for my misoperation has opened a new PR :(

jonathanslenders:
Thanks for this!

I still need to think about it, but will probably merge it one way or the other.

Some thoughts I have right now:
- Why not make "lazy" a parameter of ``FileHistory``?
- What about making ``FileHistory`` a context manager?
- I think there is some code duplication between this ``__exit__`` and the ``FileHistory`` his exit.
- Does it work correctly when wrapped in ``ThreadedHistory``?
- I'd probably like to have some good unit tests for all this, before actually merging it.

urain39:
```python
class FileHistory(History):
    """
    :class:`.History` class that stores all strings in a file.
    """
    def __init__(self, filename, lazymode=False):
        self.filename = filename
        self.lazymode = lazymode
        self._history_next_index = 0
        super(FileHistory, self).__init__()

    def __enter__(self):
        """:return: self
        """
        return self

    def __exit__(self, exc_type, exc_val, exc_tb):
        if not self.lazymode or len(self.history_strings) < 1:
            return  # Skip lazymode off and empty.

        with open(self.filename, 'ab') as f:
            lines = ['\n# {0}\n'.format(datetime.datetime.now())]

            for line in self.history_strings[self._history_next_index:]:
                if line.count('\n') > 0:
                    # multi-line
                    lines.append('\n')
                    for _line in line.split('\n'):
                        lines.append('+%s\n' % _line)
                else:
                    # single-line
                    lines.append('\n+%s\n' % line)
            # Save to file in once write.
            f.write(''.join(lines).encode('utf-8'))

    def load_history_strings(self):
        strings = []
        lines = []

        def add():
            if lines:
                # Join and drop trailing newline.
                string = ''.join(lines)[:-1]

                strings.append(string)

        if os.path.exists(self.filename):
            with open(self.filename, 'rb') as f:
                for line in f:
                    line = line.decode('utf-8')

                    if line.startswith('+'):
                        lines.append(line[1:])
                    else:
                        add()
                        lines = []

                add()

        self._history_next_index = len(strings)

        # Reverse the order, because newest items have to go first.
        return reversed(strings)

    def store_string(self, string):
        if self.lazymode:
            return  # Only saving when exiting on lazymode.

        # Save to file.
        with open(self.filename, 'ab') as f:
            def write(t):
                f.write(t.encode('utf-8'))

            write('\n# %s\n' % datetime.datetime.now())
            for line in string.split('\n'):
                write('+%s\n' % line)

    @property
    def history_strings(self):
        return self.get_strings()
```

Like this? I think `__exit__` saving to file just another way, if we let FileHistory be a contex manager(-like). And it's easy to modify the history format in one class.

urain39:
Closed, maybe someone can implement it better :)

PR #792, opened 2018-11-21
fix CommandLineInterface to Application in doc and comments

vaaaaanquish:
I think that doc and comments should also be fixed, because Application has merged CommandLineInterface.
The notation of CommandLineInterface leads to a misunderstanding of beginners.
 
Sorry about minor thing. Thanks.


jonathanslenders:
Thank you!

PR #795, opened 2018-11-26
New example for multi threading

makitos666:
New example to generate a single updated progress bar from several threads at the same time.


jonathanslenders:
Thank you!
This example made me realize that I have to improve the API of the progress bar. A couple of readable methods should do the job. I don't think I want people to call ``iter`` on a ``ProgressBar`` object.

PR #800, opened 2018-12-06
Call accept_handler before append_to_history in Buffer.validate_and_handle

asmeurer:
This was a regression from 6cfc3f3059db9816a09e2ff91a923897ceabdd10.

Without this, any changes made to the buffer text in the accept handler are
not reflected in the history. For instance, I call dedent(text).strip() in the
accept handler.


jonathanslenders:
Thank you @asmeurer, for creating this pull request! This makes sense.

PR #804, opened 2018-12-09
Add Python 3.7 testing back to Travis, and enable pip caching

orf:
3.7 is available with the `xenial` image 🎉 

~It requires `sudo`, but seeing as Travis is deprecating the `sudo: false` containers anyway it's not a big deal?~ Edit: that causes 2.7 and pypy to break. So lets just use it in the 3.7 build 👍

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,11 @@
-sudo: false
+dist: xenial
+sudo: required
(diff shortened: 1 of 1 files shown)


jonathanslenders:
Thank you @orf!

PR #805, opened 2018-12-09
Make invalidate functions get properly removed from Events.

Clairtymaker:
The remove invalidate calls don't refer to the same function as the add, so the events just keep adding calls to invalidate without removing them.


jonathanslenders:
Thank you @Clairtymaker, this was a bug indeed!
I wonder how you noticed this, actually, an "invalidate" that is called too often isn't that visible.

Clairtymaker:
I noticed that completion was getting slower after my application was running for several hours on a really old laptop. Tracked it down to invalidate being called tens of thousands of times at every key press.

PR #810, opened 2019-01-04
Change capitalization of [S]tyle.from_dict

j6k4m8:
Current docs use a lowercase `style` which is undefined in this code snippet.


jonathanslenders:
Thanks!

PR #812, opened 2019-01-06
Add litecli to Projects using prompt_toolkit

omidraha:
Add `litecli` to Projects using prompt_toolkit, and update `pgcli` and `mycli` links.


jonathanslenders:
Thanks!

PR #815, opened 2019-01-07
Fix ResourceWarning: unclosed file in setup.py

BoboTiG:
Signed-off-by: Mickaël Schoentgen <contact@tiger-222.fr>


jonathanslenders:
Thanks!

PR #817, opened 2019-01-10
Implemented a generic FuzzyCompleter that wraps around any other Completer

jonathanslenders:


asmeurer:
This is very cool. I like the formatting in the completion list. 

When I first saw this I thought "fuzzy" meant typo matching using `difflib.get_close_matches`. Any thoughts on adding something like that as well? 

Another thing to think about here is the ordering of the completions. Maybe it's out of scope, but for fuzzy completions, you generally want some kind of smart ordering. I don't know how other things do it, other than most recently used first. Maybe also prioritizing if one of the typed letters the first letter of a word separator (including CamelCase).

jonathanslenders:
Thanks! This was originally implemented as an external library made by @amjith. I believe it was tested and used for multiple projects before this integration.
I do like other approaches like difflib as well, but I think I'd like to follow the same approach. Have someone implement it as an external library, and when proven valuable for several users, merge it.

asmeurer:
I've noticed since enabling this that completions often block in a very bad way. That is to say that the prompt completely freezes up and I can't type anything until all the completions are found. I use Jedi, which sometimes is very slow to get all the completions on large libraries.

prompt_toolkit/completion/fuzzy_completer.py
```diff
@@ -65,45 +96,76 @@ def sort_key(fuzzy_match):
         fuzzy_matches = sorted(fuzzy_matches, key=sort_key)
```
asmeurer: Based on this I'm assuming that this completer wasn't designed to emit the completions as they come in. But even so it's still concerning that this blocks. I have `complete_in_thread` set to True.

PR #851, opened 2019-03-04
Advanced tempfile for Buffer/Editor

njalerikson:
`tempfile` argument for **Buffer** which allows the user to provide a partial path in order to enable more advanced syntax highlighting (e.g. git commit message highlighting).

Example code using tempfile:
```python
from prompt_toolkit import prompt

bindings = KeyBindings()


@bindings.add("c-t", "e")
def event(event):
    event.app.current_buffer.open_in_editor()


prompt("Git commit message:", tempfile=".git/COMMIT_EDITMSG", key_bindings=bindings)
```

### DIFF
--- a/prompt_toolkit/buffer.py
+++ b/prompt_toolkit/buffer.py
@@ -6,6 +6,7 @@ import asyncio
 import os
 import re
 import shlex
+import shutil
 import subprocess
 import tempfile
 from asyncio import Task, ensure_future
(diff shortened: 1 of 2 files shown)


jonathanslenders:
Hi @njalerikson,

Sorry for coming back to this PR only now.
Is it still relevant? If so, can you rebase it on the 3.0 branch? I think it does make sense.

njalerikson:
@jonathanslenders I still think this is a valuable change - I have rebased the changes and included an example code above.

prompt_toolkit/buffer.py
```diff
@@ -1398,7 +1405,31 @@ def open_in_editor(self, validate_and_handle: bool = False) -> 'asyncio.Task[Non
             raise EditReadOnlyBuffer()
 
         # Write to temporary file
-        descriptor, filename = tempfile.mkstemp(to_str(self.tempfile_suffix))
+        remove = None
```
jonathanslenders: This line needs to be removed, according to mypy. Otherwise, ``remove`` will be ``Optional[str]``, and ``rmtree`` below doesn't like that.

Can you do this, and then squash all the commits into one? Then I'll merge it.

njalerikson:
I believe that's all of it!

PR #857, opened 2019-03-13
Fix option-up and option-down on Mac with iTerm2

asmeurer:
Like option-left and -right, the key sequences start with 9. See issue #483.


jonathanslenders:
Thank you @asmeurer, I really appreciate it!

PR #859, opened 2019-03-21
Spelling: Furter > Further

PQYPLZXHGF:
Pretty sure you don't mean [Furter](https://en.wiktionary.org/wiki/furter)


jonathanslenders:
Thanks!

PR #877, opened 2019-04-13
fallback for incorrectly reported terminal size

davidszotten:
add back fallback fix from 9ad1dd9d that got lost in the refactor 6970fbda


davidszotten:
i'm not sure why it's different, but for the prompting use case, i do get the correct answer with `curses.tigetnum('cols'))`.  would there be downsides to first falling back to this before the hard coded value?

jonathanslenders:
Thanks!

davidszotten:
what are your thoughts on using curses to get this info (as fallback or instead)?

PR #878, opened 2019-04-14
Update README.rst

Zeerg:
The Ergonomica link no longer works and should be pointed to the git repo instead.


jonathanslenders:
Thanks!

PR #885, opened 2019-04-20
Drop support for Python 2

### LINKED ISSUE #801: Drop python 2
I see that you had to delete type annotations in one of the latest commits, because they are not supported in python 2, what about dropping support for it? I think it would make it much easier for maintaining this project. Big projects like django already dropped support for python 2

hugovk:
Fixes #801.


jonathanslenders:
Thank you @hugovk!

There is some overlap with the changes I do in https://github.com/prompt-toolkit/python-prompt-toolkit/pull/880  , but when I'm done there, I'll merge all of this in a prompt_toolkit 3.0 branch.

jonathanslenders:
Hi @hugovk,

In the last few weeks, I've been working on prompt_toolkit 3.0 which uses Python 3, has asyncio natively and type annotations everywhere. This includes probably most of the changes we have here. Most of the effort was writing the annotations, and while doing that, I removed Python 2 specific stuff.
This work you can now find on the master branch. We keep the 2.0 branch for updates on version 2.
Feel free to create a new pull request against this master if you notice that there's something I forgot.

I hope you don't mind that I close this.

Thanks,
Jonathan

hugovk:
👍 Thanks, that makes sense and sounds good.

PR #897, opened 2019-05-14
Add PROMPT_TOOLKIT_NO_CPR=1 environment variable to disable CPR requests (2.0)

asmeurer:
Add PROMPT_TOOLKIT_NO_CPR=1 environment variable to disable CPR requests.

See https://github.com/prompt-toolkit/python-prompt-toolkit/issues/886#issuecomment-490274914.

PR #898, opened 2019-05-14
Add PROMPT_TOOLKIT_NO_CPR=1 environment variable to disable CPR requests (master)

asmeurer:
Same as https://github.com/prompt-toolkit/python-prompt-toolkit/pull/897 execpt against master.

I've tested https://github.com/prompt-toolkit/python-prompt-toolkit/pull/897 and can confirm it works, but I have not tested the master version (this branch).


jonathanslenders:
I assume it works on master too, I'll test later. The change looks obvious.

jonathanslenders:
Thanks!

PR #899, opened 2019-05-15
Render by default to stderr if stdout is redirected.

jonathanslenders:


hotpxl:
fwiw, sometimes I want to redirect prompt_toolkit (I use IPython terminal) to a different tty, and I couldn't do that before because sys.__stdout__ couldn't be changed. But now I can do this.

Thanks for the patch!

PR #901, opened 2019-05-17
Added type annotations to PromptSession.prompt() and added prompt_asy…

jonathanslenders:
…nc() method.

### DIFF
--- a/prompt_toolkit/shortcuts/prompt.py
+++ b/prompt_toolkit/shortcuts/prompt.py
@@ -33,9 +33,11 @@ from typing import (
     Any,
     Callable,
     Generator,
+    Generic,
     List,
     Optional,
     Tuple,
+    TypeVar,
     Union,
     cast,
 )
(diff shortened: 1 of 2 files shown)


prompt_toolkit/shortcuts/prompt.py
```diff
@@ -846,11 +873,17 @@ def run_sync() -> Any:
...
+        if _async:
             return ensure_future(run_async())
         else:
             return run_sync()
 
+    async def prompt_async(self, *a: Any, **kw: Any) -> Any:
+        """
+        Like the `prompt()` method, but return a coroutine.
+        """
+        return await self.prompt(_async=True)
```
bjorn3: The `*a` and `**kw` args are not used.

jonathanslenders: Thanks for pointing out @bjorn3, these would need to be passed along to ``self.prompt``.

I'm still not sure though, whether I want to use ``*args`` and ``**kwargs``, instead of repeating all the arguments. Mypy can't infer types in case of ``*args`` and ``**kwargs`` like here.

PR #906, opened 2019-05-28
Update README readability

gpotter2:
I personally find the README too crowded. **This PR is *very* subjective: feel free to request any change until you're satisfied.**

This PR:
- moves the projects that use `prompt_toolkit` to another file. The idea is nice, but being realistic that's not what a user is looking for, and it takes tons of space.
- fix quote escapes
- Add a link to the gallery + rename "ptpython" category into "Gamlery" (with the ptpython as an example). It makes more sense of why such a category should be on top.
- remove "Runs from 2.6 to 3.7" (not sure of this one ?! I however left the Unicode comment for Python 2.7)
- Move the indtallation above Windows. It makes more sense to me to first show how easy it is to install, then show what issues there are.

**DEMO: https://github.com/gpotter2/python-prompt-toolkit/blob/gallery/README.rst**

### DIFF
--- a/PROJECTS.rst
+++ b/PROJECTS.rst
@@ -1,0 +1,61 @@
+Projects using `prompt_toolkit`
+===============================
+
+Shells:
+
+- `ptpython <http://github.com/prompt-toolkit/ptpython/>`_: Python REPL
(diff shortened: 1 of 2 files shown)


jonathanslenders:
This looks very good. Thanks!

Issue #907, opened 2019-05-30
button_dialog example AttributeError


OphirCarmi:
Traceback (most recent call last):
  File "scratch_46.py", line 23, in <module>
    main()
  File "scratch_46.py", line 15, in main
    ('Maybe...', None),
AttributeError: 'bool' object has no attribute 'run'

need to remove ".run()"

jonathanslenders:
Hi @OphirCarmi,

As pointed out in the PR, the ``run()`` call is only required in prompt_toolkit 3.0, which is not yet released. Check the 2.0 branch for examples for prompt_toolkit 2.0.

OphirCarmi:
Ok, thanks for your quick reply.

PR #908, opened 2019-05-30
Update button_dialog.py

OphirCarmi:
#907


jonathanslenders:
Hi @OphirCarmi,

The original code was fine. You're looking at the master branch, which is prompt_toolkit 3.0. This was one of the few breaking changes we had to do (regarding the asyncio changes).

Check the 2.0 branch for the examples of the latest released version.

Jonathan

PR #911, opened 2019-06-02
yes_no_dialog doesn't require a run call

mikaeilorfanian:


jonathanslenders:
Hi @mikaeilorfanian,

Thanks for the PR. However, it does require a run call on the master branch (which is prompt_toolkit 3.0). Check the 2.0 branch for the 2.0 examples.

PR #914, opened 2019-06-06
removed extra  "the"

PQYPLZXHGF:


docs/pages/upgrading/3.0.rst
```diff
@@ -88,7 +88,7 @@ application (within a coroutine), because it will try to run the event loop
 again. In that case, always use `prompt_async()`.
 
 
-Changes to the the dialog functions
+Changes to the dialog functions
 -----------------------------------
```
gpotter2: You would need to resize the separator then

PQYPLZXHGF: Done

PR #918, opened 2019-06-10
Allow PromptSession to take Document as `default`

pritambaral:
Text defaults in prompts are nice, but sometimes it is nicer to be able
to specify more than just the text in the default.

For example, I had a need to prompt for a value with a suggested suffix. With the normal `prompt(default=suffix)`, the user would have to move the cursor to the beginning of the prompt. But with `prompt(default=Document(text=suffix, cursor_position=0))` the hassle is avoided.

Because `PromptSession` converts the provided default into a
`Document` anyway, this change makes `PromptSession` also accept a pre-made
Document.


jonathanslenders:
This makes sense indeed. Thanks for the pull request!

Issue #915, opened 2019-06-06
assertion in print_formatted_text


jnwatson:
prompt toolkit version 2.0.9

Example code:
```
from prompt_toolkit import print_formatted_text
print_formatted_text('Hello world\r')
```
Output:
```
Traceback (most recent call last):
  File "/tmp/pttest.py", line 3, in <module>
    print_formatted_text('Hello world\r')
  File "/home/nic/.pyenv/versions/3.7.1/lib/python3.7/site-packages/prompt_toolkit/shortcuts/utils.py", line 124, in print_formatted_text
    style_transformation=style_transformation)
  File "/home/nic/.pyenv/versions/3.7.1/lib/python3.7/site-packages/prompt_toolkit/renderer.py", line 674, in print_formatted_text
    assert '\r' not in text
AssertionError
```
I might suggest that an assertion isn't appropriate here, given that `print_formatted_text` is used as a substitute for `print`

jonathanslenders:
Hi @jnwatson,

If you have some time, feel free to create a pull request here that removes carriage returns. (Exactly like done in your project.)
This can then be merged in both the 2.0 and master branch.

PR #922, opened 2019-06-12
Remove no-\r assertion in print_formatted_text

jnwatson:
Follow up from https://github.com/prompt-toolkit/python-prompt-toolkit/issues/915

I'll go ahead and submit a similar PR for 2.0 branch once this one's accepted.

### DIFF
--- a/prompt_toolkit/renderer.py
+++ b/prompt_toolkit/renderer.py
@@ -684,7 +684,7 @@ def print_formatted_text(
         else:
             output.reset_attributes()
 
-        # Eliminate newlines
+        # Eliminate carriage returns
         text = text.replace('\r', '')
 
         # Assume that the output is raw, and insert a carriage return before
--- a/prompt_toolkit/renderer.py
+++ b/prompt_toolkit/renderer.py
@@ -684,9 +684,11 @@ def print_formatted_text(
         else:
             output.reset_attributes()
 
+        # Eliminate newlines
+        text = text.replace('\r', '')
+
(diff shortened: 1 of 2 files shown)


prompt_toolkit/renderer.py
```diff
@@ -684,9 +684,11 @@ def print_formatted_text(
         else:
             output.reset_attributes()
 
+        # Eliminate newlines
+        text = text.replace('\r', '')
```
bjorn3: This would incorrectly remove bare `\r` (eg old mac newline)

jnwatson:
As far as I know, prompt-toolkit doesn't run on pre-OS-X macs. The latest
Python distribution I could find for such a platform was 2.3.3.

prompt_toolkit/renderer.py
```diff
@@ -684,9 +684,11 @@ def print_formatted_text(
         else:
             output.reset_attributes()
 
+        # Eliminate newlines
```
jonathanslenders: Technically, these are actually the "carriage returns" which we are eliminating. Would you mind rephrasing the comment? Then it's perfect.

jnwatson: ok

jonathanslenders:
Thanks!

jonathanslenders:
And thanks for the unit test too! :-)

jonathanslenders:
I just cherry-picked this commit on the 2.0 branch too, so no need to create a second PR.

jnwatson:
My pleasure. Thank you for all of your hard work.

PR #930, opened 2019-06-21
Make sure HANDLE  on Windows has a correct size

quark-zju:
Previously, the type of various HANDLEs are native Python integer types.
The ctypes library will treat them as 4-byte integer when used in function
arguments. On 64-bit Windows, HANDLE is 8-byte and usually a small integer.
Depending on whether the extra 4 bytes are zero-ed out or not, things can
happen to work, or break.

This patch adds type casting so ctypes uses 8-byte integers for HANDLEs
on 64-bit Windows.

Fixes #406.

This pull request is based on the 1.0 branch, since that's used by the
last IPython release for Python 2 and we're currently using IPython
+python2 (64bit MSVC2017 build).

PR #938, opened 2019-07-16
Make sure HANDLE on Windows has correct size

jonathanslenders:
Previously, the type of various HANDLEs are native Python integer types.
The ctypes library will treat them as 4-byte integer when used in function
arguments. On 64-bit Windows, HANDLE is 8-byte and usually a small integer.
Depending on whether the extra 4 bytes are zero-ed out or not, things can
happen to work, or break.

This patch adds explicit type casts so ctypes uses 8-byte integers for
HANDLEs on 64-bit Windows.

Thanks to @quark-zju for providing this patch to the 1.0 branch. This patch
contains similar changes for the the 2.0 branch.


jonathanslenders:
See also: https://github.com/prompt-toolkit/python-prompt-toolkit/pull/930

@quark-zju: These are your changes applied to the 2.0 branch. Feel free to have a look if you have time. I did not yet try it.

PR #944, opened 2019-07-29
The "migrating to 2.0" documentation should menion that the word_completer directory moved

ReddyKilowatt:
In prompt_toolkit 2.0, the word_completer is now located in prompt_toolkit\completion, instead of under the contribs directory. It would be helpful if the documentation mentions the new location.


jonathanslenders:
Closing this.

This was supposed to be an issue.
Also see: https://github.com/prompt-toolkit/python-prompt-toolkit/pull/947

PR #950, opened 2019-08-02
Added stop_time to ProgressBarCounter

njalerikson:
When using ProgressBars it makes sense for elapsed time to stop incrementing once the counter in question finishes.

### DIFF
--- a/prompt_toolkit/shortcuts/progress_bar/base.py
+++ b/prompt_toolkit/shortcuts/progress_bar/base.py
@@ -307,6 +307,7 @@ class ProgressBarCounter(Generic[_CounterItem]):
                  total: Optional[int] = None) -> None:
 
         self.start_time = datetime.datetime.now()
+        self.stop_time = None
         self.progress_bar = progress_bar
         self.data = data
         self.items_completed = 0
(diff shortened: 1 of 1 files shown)


prompt_toolkit/shortcuts/progress_bar/base.py
```diff
@@ -306,6 +306,7 @@ def __init__(self, progress_bar: ProgressBar,
                  total: Optional[int] = None) -> None:
 
         self.start_time = datetime.datetime.now()
+        self.stop_time = None
```
jonathanslenders: I think this requires a type annotation, otherwise, Mypy will not allow assigning anything else but ``None`` to ``stop_time``.
```python
self.stop_time: Optional[datetime.datetime] = None
```

prompt_toolkit/shortcuts/progress_bar/base.py
```diff
@@ -348,6 +349,10 @@ def done(self) -> bool:
     def done(self, value: bool) -> None:
         self._done = value
 
+        # if declared done then store the stop_time
```
jonathanslenders: If you don't mind, start comments with a capital and end with a period.

jonathanslenders:
Thanks @njalerikson!

Issue #909, opened 2019-05-30
Long Names


mlucool:
Hi,

If I have a long name I am trying to autocomplete, this no longer is useful. Would it be possible to allow for one of the following:
1. Allow the part already typed to not be included
2. Ellipsis the start not the end of the string
3. Allow for the text displayed in the UI to not be the text updates what is selected

While below is contrived, this happens with autocomplete of long chains of properties. Reproducer:
```
import random
import string

def randomString(stringLength=10):
    """Generate a random string of fixed length """
    letters = string.ascii_lowercase
    return ''.join(random.choice(letters) for i in range(stringLength))

def my_completer(c, text):
    longstr = 'AUTO_COMPLETE_A_VERY_LONG_STRING_BECAUSE_SOMETIMES_THIS_HAPPENS'
    if text.startswith(longstr):
        return [(longstr + randomString()).decode('utf8') for x in range(150)]
    return []

get_ipython().set_custom_completer(my_completer, 0)
```
![image](https://user-images.githubusercontent.com/1813603/58670975-ee29ac00-830e-11e9-9fbc-67ddeb6a84ad.png)

Python 2.7.15
IPython 5.8.0
PromptToolkit: 1.0.15

Please also let me know if this is available in a later version of prompt_toolkit.

jonathanslenders:
Hi @mlucool,

This totally makes sense to have, but is unfortunately not yet available in prompt_toolkit.
It can be done though, let's take this as a feature request.

utabe:
@jonathanslenders  I'm interesting in this as well. Where in the code would a change need to be made to make this happen?

utabe:
Would modifying `_trim_formatted_text` in prompt_toolkit\layout\menus.py work as a solution to option 2 listed above? 

> Ellipsis the start not the end of the string

quarl:
(The original https://github.com/prompt-toolkit/python-prompt-toolkit/issues/909#issue-450548763 May30 request came from me.)

>     1. Allow the part already typed to not be included

Clarifying the above approach1: 
Suppose:
- the input text is: foo1.foo2.foo3.foo4.foo5.foo6.bar<TAB>
- foo6's completions are bar1, bar2, bar3

[...]

PR #954, opened 2019-08-12
Add ... at start instead of end of trimmed formatted text

utabe:
Add support for issue Long names 909.

### DIFF
--- a/prompt_toolkit/layout/menus.py
+++ b/prompt_toolkit/layout/menus.py
@@ -223,17 +223,23 @@ def _trim_formatted_text(
     if width > max_width:
         result = []  # Text fragments.
         remaining_width = max_width - 3
+        current_width = 0
 
         for style_and_ch in explode_text_fragments(formatted_text):
             ch_width = get_cwidth(style_and_ch[1])
 
+            # Add '...' somewhere in the middle of the text
+            if current_width == max_width - 6:
+                result.append(('', '...'))
+
             if ch_width <= remaining_width:
                 result.append(style_and_ch)
                 remaining_width -= ch_width
+                current_width += ch_width
             else:
                 break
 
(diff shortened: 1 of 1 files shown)


utabe:
@jonathanslenders what would need to be added to this in order for it to mergeable?

mlucool:
@jonathanslenders Any objection to this request?

mlucool:
@jonathanslenders just wanted to reach out again to see your thoughts on this

ngoldbaum:
@jonathanslenders any chance this one can get some code review?

ngoldbaum:
I guess this one doesn't have a ton of context, I'm going to open a new pull request that will hopefully be easier to digest for the maintainers.

asmeurer:
I just tested this and it works. It makes more sense for ... to go at the beginning because the beginning of the completion is already there in the text that was typed.

jonathanslenders:
Thanks for this! (And thanks for the patience, it's been super busy.)

I think usually it does make sense indeed, but not always. There are situations where we have to complete long filenames with many pieces in common, where nothing was typed yet. for instance:
```
____________a__________1_________
____________a__________2_________
____________b__________1_________
____________b__________2_________
etc...
```

People type this by tabbing the common pieces and typing the things that are different. If we truncate on the left, this won't work anymore. So, this will break some people's workflow.

I'll see if it's feasible to have dots on the left if the start of the completion was already typed. If too difficult, I should probably merge this.

asmeurer:
Another option might be to show both the part right after what is typed as well as the end, and have the ellipses in the middle. I don't know how hard that is to do.

jonathanslenders:
Related issue: https://github.com/prompt-toolkit/python-prompt-toolkit/issues/909

PR #962, opened 2019-09-04
Doc: fix example code

Hanaasagi:
`EditingMode` is not imported in example code.


jonathanslenders:
Thank you @Hanaasagi!

PR #970, opened 2019-09-25
Specify attribute name for Label

njalerikson:
Instead of only displaying the `label` attribute of a `ProgressBarCounter` we allow the `Label` formatter to specify which attribute name to process.

@jonathanslenders what are your thoughts on also changing `ProgressBarCounter` into this:

```python
def __init__(self, progress_bar: ProgressBar,
             data: Optional[Iterable[_CounterItem]] = None,
             # label: AnyFormattedText = '',  # remove this
             remove_when_done: bool = False,
             total: Optional[int] = None,
             **labels: AnyFormattedText,  # add this
             ) -> None:
```

The above change would break backwards compatibility so I'm hesitating. Alternatively, we could do something like this:

```python
def __init__(self, progress_bar: ProgressBar,
             data: Optional[Iterable[_CounterItem]] = None,
             label: AnyFormattedText = '',
             remove_when_done: bool = False,
             total: Optional[int] = None,
             **extra_labels: AnyFormattedText,  # add this
             ) -> None:
```


jonathanslenders:
Hi @njalerikson,

I think I understand what you're trying to do, but I don't really like the use of kwargs and AttributeError. (Catching AttributeError will hide too many bugs by accident, and the usage of kwargs make it hard to introduce new keyword arguments).

So, if we need to pass extra data along with every ``ProgressBarCounter``, let's call it ``extra_dict``. Just a free-form dictionary, ``Dict[object, object]`` really. Then maybe create a ``CustomFormatter`` that takes any callable like ``Callable[['ProgressBarCounter[object]'], AnyFormattedText]``.

Then ``Label`` can be implemented as a specialization of this ``CustomFormatter``.

```python
class Label(CustomFormatter):
    def __init__(self, ...) -> None:
        super().__init__(lambda progress: progress.label)
```
(something like that - I'm skipping over some details.)

jonathanslenders:
Assigning an ``action`` attribute of a ``ProgressBarCounter`` instance, also violates what is possible with static type checking. Either we use a dictionary, or we allow users to inherit from ``ProgressBarCounter`` and define their own fields.

Maybe we should go for the dictionary approach, and take both ``label`` and ``labels`` as input, where the singular ``label`` stores its value under the ``"default"`` key of the ``labels`` dictionary.
I guess then it becomes a ``Dict[str, str]``.

njalerikson:
Added the labels argument and removed the AttributeError catching.

Anything else that we'd like to change here?

prompt_toolkit/shortcuts/progress_bar/base.py
```diff
@@ -313,6 +317,8 @@ def __init__(self, progress_bar: ProgressBar,
         self.remove_when_done = remove_when_done
         self._done = False
         self.total: Optional[int]
+        for name, label in labels.items():
+            setattr(self, name, label)
```
jonathanslenders: At this point, I prefer not to use ``setattr``, but literally assign it to ``self.labels``. ``label`` can also be added there. Something like this:
```python
self.labels = {"label": label, ** labels}
```

prompt_toolkit/shortcuts/progress_bar/formatters.py
```diff
@@ -76,7 +77,9 @@ class Label(Formatter):
     :param suffix: String suffix to be added after the task name, e.g. ': '.
         If no task name was given, no suffix will be added.
     """
-    def __init__(self, width: AnyDimension = None, suffix: str = '') -> None:
+    def __init__(self, width: AnyDimension = None, suffix: str = '', *,
+                 name: Optional[str] = None) -> None:
+        self.name = operator.attrgetter(name or self.__class__.__name__.lower())
```
jonathanslenders: This line can then be:
```python
self.name = name
```
(I prefer that attributes, received in ``__init__`` are assigned to self, unmodified if possible. I know that we are violating this sometimes with the ``to_filter`` stuff, but I think there's no need here.)

prompt_toolkit/shortcuts/progress_bar/formatters.py
```diff
@@ -88,7 +91,7 @@ def format(self, progress_bar: 'ProgressBar',
                progress: 'ProgressBarCounter[object]',
                width: int) -> AnyFormattedText:
 
-        label = self._add_suffix(progress.label)
+        label = self._add_suffix(self.name(progress))
```
jonathanslenders: I think:
```python
label = self._add_suffix(self.progress.labels.get(self.name, ''))
```

jonathanslenders:
Thanks for the changes @njalerikson! The few comments I had were mostly about certain things that can't be typechecked with mypy (like setattr and attrgetter).

prompt_toolkit/shortcuts/progress_bar/base.py
njalerikson: Just to confirm this will mean that to update the labels during runtime the user will end up doing this:

```python
bar.label = "new label"
bar.labels["action"] = "new value"
```

If that's what we want then I'll make these changes.

njalerikson: So inject `label` into `labels` in `__init__` and then provide a getter/setter `label` for backwards compatibility?

njalerikson: oh I just processed the example you had in the first comment, yes we are on the same page

njalerikson:
@jonathanslenders where are we on this?

jonathanslenders:
This looks good to me. Sorry for the late reply! (I've been busy.)

Can you squash the four commits into one, and maybe remove the unused "operator" import? Then I'll merge it!

njalerikson:
no worries - heck it's taken me 2 weeks to get back to this...

njalerikson:
Ended up remove several unused imports, thanks for catching that. Anything else this needs?

njalerikson:
@jonathanslenders what's needed here? anything I should modify?

PR #974, opened 2019-09-27
ProgressBarCounter delayed removal

njalerikson:
@jonathanslenders - this is dependant on #950

### DIFF
--- a/prompt_toolkit/shortcuts/progress_bar/base.py
+++ b/prompt_toolkit/shortcuts/progress_bar/base.py
@@ -32,6 +32,7 @@ from typing import (
     Sized,
     TextIO,
     TypeVar,
+    Union,
     cast,
 )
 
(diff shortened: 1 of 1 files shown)


jonathanslenders:
Can you explain a bit why exactly this feature would be good to have?
I have some comments about the implementation, but first, I'd like to understand the use case. 

(It sounds like a very specific use case, and in that case, it shouldn't be in prompt_toolkit itself, but we should expose the APIs to make it possible to build this on top.)

njalerikson:
I have developed some relatively complex progress bars that contain a lot of information. For example, I have one example that does bulk downloads of files and upon downloading run some tests/post download process. In another example, I'm running a bulk SVN to Git migration where after cloning the SVN repo additional processing is done to clean up the repo prior to pushing. In other words, both of these processes contains a lot of different information. And when a process succeeds (or fails) I want to have an opportunity to inspect the bar prior to the bar being removed - hence a delayed remove.

I can agree that this feature could be better implemented as some kind of callable - that way it could be a conditional remove (e.g. only keep the last 5 bars or a time delay removal).

njalerikson:
closing in favor of #1025

PR #975, opened 2019-09-27
ProgressBar Counter formatter

njalerikson:
Counter that pads itself based on the total value.


prompt_toolkit/shortcuts/progress_bar/formatters.py
```diff
@@ -198,6 +199,29 @@ def get_width(self, progress_bar: 'ProgressBar') -> AnyDimension:
         return D.exact(max(all_lengths) * 2 + 1)
 
 
+class Counter(Formatter):
```
jonathanslenders: In order to avoid duplication, can you create a similar ``Formatter`` for ``Total``, and then rewrite the ``Progress`` formatter so that it uses the other two underneath?

There is a ``merge_formatted_text`` utility to concatenate ``HTML`` objects.

prompt_toolkit/shortcuts/progress_bar/formatters.py
```diff
@@ -198,6 +199,29 @@ def get_width(self, progress_bar: 'ProgressBar') -> AnyDimension:
...
+    """
+    Display the items competed as text.  E.g. "8"
```
jonathanslenders: Can you add ``Counter`` to ``__all__`` at the top of this file as well?

njalerikson: done

prompt_toolkit/shortcuts/progress_bar/formatters.py
njalerikson: done

njalerikson:
Cleaning up old PRs

Issue #976, opened 2019-09-29
Unhandled exception in event loop in auto_suggest keybindings


vaaaaanquish:
Hi. Thanks.

# What
I got `Unhandled exception in event loop` error in this keybinding function.
https://github.com/prompt-toolkit/python-prompt-toolkit/blob/master/prompt_toolkit/key_binding/bindings/auto_suggest.py#L49

# Environment
python 3.6.8
prompt-toolkit == 2.0.9 or 3.0.0

# How Reproduction
using this sample: https://github.com/prompt-toolkit/python-prompt-toolkit/blob/master/examples/prompts/auto-suggestion.py
```
Say something: import [[[[escape f]]] path [[[[escape f]]] 

Say something: import path/Users/shunsuke.kawai/homebrew/versions/3.6.8/lib/python3.6/site-packages/prompt_toolkit/key_binding/key_processor.py:273: DeprecationWarning: generator 'KeyProcessor._process' raised StopIteration

Unhandled exception in event loop:
  File ".../prompt_toolkit/eventloop/posix.py", line 154, in _run_task
  File ".../prompt_toolkit/eventloop/context.py", line 115, in new_func
  File ".../prompt_toolkit/application/application.py", line 562, in read_from_input
  File ".../prompt_toolkit/key_binding/key_processor.py", line 273, in process_keys
  File ".../prompt_toolkit/key_binding/key_processor.py", line 180, in _process
  File ".../prompt_toolkit/key_binding/key_processor.py", line 323, in _call_handler
  File ".../prompt_toolkit/key_binding/key_bindings.py", line 78, in call
  File ".../prompt_toolkit/key_binding/bindings/auto_suggest.py", line 51, in _

Exception
Press ENTER to continue...
```

# Solution

Because generator raises `StopIteration`.
We use suggestion_available filter, but `app.current_buffer.suggestion` becomes None or
`Suggestion()` .
https://github.com/prompt-toolkit/python-prompt-toolkit/blob/master/prompt_toolkit/key_binding/bindings/auto_suggest.py#L31

We can use `try/except` or `Suggestion.__len__`.

## A: try/except
https://github.com/prompt-toolkit/python-prompt-toolkit/blob/master/prompt_toolkit/key_binding/bindings/auto_suggest.py#L49
```
    @handle('escape', 'f', filter=suggestion_available & emacs_mode)
    def _(event: E) -> None:
        b = event.current_buffer
        suggestion = b.suggestion
[...]

PR #977, opened 2019-09-29
fix auto_suggest keybindings

vaaaaanquish:
fix: https://github.com/prompt-toolkit/python-prompt-toolkit/issues/976


prompt_toolkit/key_binding/bindings/auto_suggest.py
```diff
@@ -30,7 +30,7 @@ def load_auto_suggest_bindings() -> KeyBindings:
     @Condition
     def suggestion_available() -> bool:
         app = get_app()
-        return (app.current_buffer.suggestion is not None and
+        return (app.current_buffer.suggestion and
```
jonathanslenders: I would probably prefer here to check against ``len(suggestion.text)``, instead of adding the ``__len__`` above. I think it's not semantically correct for ``Suggestion`` to have a length, it's not a container type.

vaaaaanquish: @jonathanslenders Thanks. I thought so. I'll try it.

vaaaaanquish:
@jonathanslenders Fixed. Please check PR :)
I will make PR for 2.0 soo.

jonathanslenders:
Thank you @vaaaaanquish!

Can you squash the commits and rebase on the current master? I can't merge it now.

jonathanslenders:
Rebasing and merging from here: https://github.com/prompt-toolkit/python-prompt-toolkit/pull/1202

PR #978, opened 2019-10-01
Two minor tweaks to widgets/base

gpotter2:
This PR:
- updates GOTO-key in `_DialogList`
- implements the `to_str` util 
- removes a leftover class tag

About the first point, creating a `CheckBoxList` with content that is formatted (for instance that has `HTML()` in it) will make it crash. This made those lists very unstable (any key would crash)...

It was a rather low-effort implementation anyways: it should be reimplemented using a fancy search toolbar.

### DIFF
--- a/prompt_toolkit/formatted_text/utils.py
+++ b/prompt_toolkit/formatted_text/utils.py
@@ -4,17 +4,28 @@ Utilities for manipulating formatted text.
 When ``to_formatted_text`` has been called, we get a list of ``(style, text)``
 tuples. This file contains functions for manipulating such a list.
 """
-from typing import Iterable, cast
+from typing import (
+    cast,
+    Callable,
+    Iterable,
+    Union
+)
 
 from prompt_toolkit.utils import get_cwidth
 
-from .base import OneStyleAndTextTuple, StyleAndTextTuples
+from .base import (
+    AnyFormattedText,
+    OneStyleAndTextTuple,
+    StyleAndTextTuples,
+    to_formatted_text,
+)
 
 __all__ = [
     "fragment_list_len",
     "fragment_list_width",
     "fragment_list_to_text",
     "split_lines",
+    "to_str",
 ]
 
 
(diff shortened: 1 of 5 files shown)


prompt_toolkit/widgets/base.py
```diff
@@ -636,14 +636,6 @@ def _(event: E) -> None:
         def _(event: E) -> None:
             self._handle_enter()
 
-        @kb.add(Keys.Any)
-        def _(event: E) -> None:
-            # We first check values after the selected value, then all values.
-            for value in self.values[self._selected_index + 1:] + self.values:
-                if value[1].startswith(event.data):
```
jonathanslenders: Instead of removing this key binding, can you replace the above ``startswith`` with:
```python
from prompt_toolkit.formatted_text import fragment_list_to_text, to_formatted_text

if fragment_list_to_text(to_formatted_text(value[1])).startswith(event.data):
```

I think that should fix the issue for now. (Maybe we could use a ``to_str`` helper in the ``formatted_text`` module.

jonathanslenders:
I'm not sure how easy it is to make the ``SearchToolbar`` work hand in hand with this control. It was mostly optimized for searching through a ``BufferControl``, not just any control. That's why I propose to keep it with the above change.

prompt_toolkit/widgets/base.py
gpotter2: Good call.

prompt_toolkit/formatted_text/utils.py
```diff
@@ -80,3 +86,17 @@ def split_lines(fragments: StyleAndTextTuples) -> Iterable[StyleAndTextTuples]:
     # line is yielded. (Otherwise, there's no way to differentiate between the
     # cases where `fragments` does and doesn't end with a newline.)
     yield line
+
+
+def to_str(value: AnyFormattedText) -> str:
```
jonathanslenders: There is already a function ``to_str`` in ``prompt_toolkit.utils``. I think we should probably merge this functionality in there.

gpotter2:
I'm sorry for the wait, I forgot about this PR.

I merged the implementations, however not in the order you would expect: To keep things very clean, I want to keep `AnyFormattedText` as the type hint. However it's not possible to import it within the `prompt_toolkit.utils` module due to import loops. Therefore I made the latter point to the new one.

gpotter2:
@jonathanslenders Hi. I rebased this PR.
We're still bumping into this issue

gpotter2:
Closing for inactivity

PR #983, opened 2019-10-11
use the right global value for current_path.

subuTestGH:
There's a bug here causing formatting to not update based on file open.


jonathanslenders:
Thanks! Good catch!

PR #987, opened 2019-10-21
Can't join MouseEventType with strings

subuTestGH:
add cast. Else causes exceptions on mouse actions.


jonathanslenders:
Thanks for this!

Issue #493, opened 2017-03-26
Possible to have different continuation prompt for hard and soft line wrap


asmeurer:
It would be nice if `get_continuation_prompt` had enough information passed to it to have a different continuation prompt depending on if the line is hard wrapped or soft wrapped. As far as I can tell this is currently impossible (from the [logic here](https://github.com/jonathanslenders/python-prompt-toolkit/blob/9b8b21c5fc50987162a3278b55663f3425013a32/prompt_toolkit/layout/margins.py#L237), the `get_continuation_prompt` has no distinction between the different calls. Even if it were just passed `n` (for the `n`-th continuation prompt) I think this would be possible.

jonathanslenders:
This is correct, and this feature request makes sense.
In ptpython, this is possible because it uses a more low-level API (which is not guaranteed to remain stable).

I take this as a feature request.

asmeurer:
I didn't know it was possible in ptpython. I'll take a look at how it works there.

asmeurer:
Actually, what I *really* want is no continuation prompt for soft-wrapped text. Ideally, the text wouldn't even soft wrap, so that if I select it and copy it, there are no newlines. Is this possible? It works this way in readline shells.

jonathanslenders:
For a single-line input, it is already the case that there is no continuation prompt.

About having one for hard line wraps, but no for soft wraps: this is not possible right now. I'm not sure what it would take to get this to work, but I'll keep it in mind.

asmeurer:
> For a single-line input, it is already the case that there is no continuation prompt.

Can you clarify what you mean here? I have a continuation prompt set up and it always shows when a single line of input fills the terminal width. If I could disable it for this case, even with hard wrapping, that would be a good start.

asmeurer:
I can get roughly the behavior I want by setting `wrap_lines` to False, with the following diff

```diff
diff --git a/prompt_toolkit/renderer.py b/prompt_toolkit/renderer.py
index 7a8fde5..a2d0d57 100644
--- a/prompt_toolkit/renderer.py
+++ b/prompt_toolkit/renderer.py
@@ -118,7 +118,8 @@ def _output_screen_diff(output, screen, current_pos, previous_screen=None, last_
     # background threads, and it's hard for debugging if their output is not
     # wrapped.)
     if not previous_screen or not use_alternate_screen:
-        output.disable_autowrap()
+        pass
+        # output.disable_autowrap()

[...]

PR #999, opened 2019-11-10
Allow empty continuation

CouleeApps:
If continuation is an empty string, instead of ignoring it and aligning with spaces, honor the non-None but empty value and don't align to the prompt width with spaces.

This is helpful if you want to configure your prompt to allow multiline editing and also want to copy the text directly from your terminal.

### DIFF
--- a/prompt_toolkit/shortcuts/prompt.py
+++ b/prompt_toolkit/shortcuts/prompt.py
@@ -1117,7 +1117,7 @@ class PromptSession(Generic[_T]):
 
         # When the continuation prompt is not given, choose the same width as
         # the actual prompt.
-        if not continuation and is_true(self.multiline):
+        if not continuation and continuation != '' and is_true(self.multiline):
             continuation = ' ' * width
 
         return to_formatted_text(
--- a/prompt_toolkit/shortcuts/prompt.py
+++ b/prompt_toolkit/shortcuts/prompt.py
@@ -1117,7 +1117,7 @@ class PromptSession(Generic[_T]):
 
         # When the continuation prompt is not given, choose the same width as
         # the actual prompt.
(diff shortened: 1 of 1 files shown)


prompt_toolkit/shortcuts/prompt.py
```diff
@@ -1117,7 +1117,7 @@ def _get_continuation(
 
         # When the continuation prompt is not given, choose the same width as
         # the actual prompt.
-        if not continuation and is_true(self.multiline):
+        if not continuation and continuation != '' and is_true(self.multiline):
```
jonathanslenders: This makes sense, but I think I prefer
```python
if continuation is not None and is_true(self.multiline):
```
Because ``None`` is the default here, and for cases where we receive some formatted text, like ``HTML()`` ``ANSI()`` or even ``[]`` (empty list), we want to have an empty continuation. So, not just for an empty string.

CouleeApps: Sounds good to me. I wasn't sure if anywhere used False or some other falsy value for this so I opted to just check empty string. If that's not the case then checking for None seems like the best solution.

jonathanslenders:
Merged here: https://github.com/prompt-toolkit/python-prompt-toolkit/commit/fc86ff03fe4358b9ff4e85dc69662e10ccd38b0a

I had to do one more little change: the default value for ``prompt_continuation`` should be ``None`` and this becomes an Optional.

asmeurer:
Copy-pasting is still problematic because of https://github.com/prompt-toolkit/python-prompt-toolkit/issues/493.

CouleeApps:
Thanks for the merge! While this doesn't fix copy paste it is at least fixing one of the problems I was having with it.

PR #1001, opened 2019-11-11
Unpack tempfile's callables

### LINKED ISSUE #1000: Issue with open_in_editor
Seems like #851 didn't properly address tempfile/tempfile_suffix being dynamic/lambda.

https://github.com/prompt-toolkit/python-prompt-toolkit/blob/c8f5b5cd6048935c64ba451443af7962ad5d9cc2/prompt_toolkit/buffer.py#L1407-L1430

Currently, it's always ending up in the first clause since PromptSession passes the values as `lambda: self.tempfile`.

https://github.com/prompt-toolkit/python-prompt-toolkit/blob/c8f5b5cd6048935c64ba451443af7962ad5d9cc2/prompt_toolkit/shortcuts/prompt.py#L467-L487

njalerikson:
Need to unpack tempfile and tempfile_suffix callables before deciding whether creating a simple or advanced temp file for the editor.

Resolves #1000

### DIFF
--- a/prompt_toolkit/buffer.py
+++ b/prompt_toolkit/buffer.py
@@ -43,7 +43,7 @@ from .filters import FilterOrBool, to_filter
 from .history import History, InMemoryHistory
 from .search import SearchDirection, SearchState
 from .selection import PasteMode, SelectionState, SelectionType
-from .utils import Event, to_str
+from .utils import Event, to_str, to_type
 from .validation import ValidationError, Validator
 
 __all__ = [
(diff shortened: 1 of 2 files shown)


prompt_toolkit/buffer.py
```diff
@@ -1405,14 +1405,40 @@ def open_in_editor(self, validate_and_handle: bool = False) -> 'asyncio.Task[Non
             raise EditReadOnlyBuffer()
 
         # Write to temporary file
-        if self.tempfile:
+        try:
```
jonathanslenders: This function is becoming long. Is there any way to break it up in maybe two or three functions. (It is hard to review).

prompt_toolkit/utils.py
```diff
@@ -261,6 +262,14 @@ def take_using_weights(
         i += 1
 
 
+def to_type(value: Union[Callable[[], _T], _T]) -> _T:
```
jonathanslenders: Can you rename it to something like ``call_if_callable``. I think the naming is a little confusing.

njalerikson: Renamed

prompt_toolkit/buffer.py
njalerikson: Split tempfile creation into simple vs complex.

jonathanslenders:
Thanks! Merged in: https://github.com/prompt-toolkit/python-prompt-toolkit/commit/c7eb02c50c384e6ab2c03225e5c76a2242a2752e

jonathanslenders:
@njalerikson : Can you please have a look at this PR: https://github.com/prompt-toolkit/python-prompt-toolkit/pull/1007 
This contains a few improvements to this code.

There is also this line:
```python
 suffix = str(suffix) if suffix else None
```
for which I don't understand the purpose. Can you please explain what this is doing?

njalerikson:
Ah I do like that change, much cleaner, this was one of the parts I wasn’t 100% happy with.

The suffix casting to string is likely unnecessary. It’s just catching any cases where the user passes some non-string (or non-None), I have no real opinion about it.

PR #1005, opened 2019-11-24
[WIP] Use asyncio.run().

jonathanslenders:
This is not correct yet, check the persistent-history example:
```
 $ python examples/prompts/history/persistent-history.py
Say something: oaeu
You said: oaeu
  File "examples/prompts/history/persistent-history.py", line 25, in <module>
    main()
  File "examples/prompts/history/persistent-history.py", line 20, in main
    text = session.prompt('Say something: ')
  File "/home/jonathan/git/python-prompt-toolkit/prompt_toolkit/shortcuts/prompt.py", line 925, in prompt
    return self.app.run()
  File "/home/jonathan/git/python-prompt-toolkit/prompt_toolkit/application/application.py", line 754, in run
    loop = get_event_loop()
  File "/usr/lib/python3.7/asyncio/events.py", line 644, in get_event_loop
    % threading.current_thread().name)
RuntimeError: There is no current event loop in thread 'MainThread'.
```

Also, this doesn't work if new buffers are attached dynamically to the layout.

### DIFF
--- a/examples/prompts/asyncio-prompt.py
+++ b/examples/prompts/asyncio-prompt.py
@@ -49,16 +49,17 @@ async def interactive_shell():
             return
 
 
-def main():
+async def main():
     with patch_stdout():
-        shell_task = asyncio.ensure_future(interactive_shell())
-        background_task = asyncio.gather(print_counter(), return_exceptions=True)
-
-        loop.run_until_complete(shell_task)
+        background_task = asyncio.ensure_future(print_counter())
+        await interactive_shell()
         background_task.cancel()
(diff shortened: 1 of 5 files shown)


jonathanslenders:
Closing this. It's probably not the right thing to do.

``asyncio.run()`` is supposed to be called in in the main function of the application. A lot of stuff can happen before ``Application.run()`` is called. If we only call ``asyncio.run()`` in ``Application.run()``, then nothing before this can schedule things in the same event loop, because ``asyncio.run()`` creates a new event loop.

PR #1010, opened 2019-11-27
[wip] Refactoring to the Keys class.

jonathanslenders:


prompt_toolkit/keys.py
```diff
@@ -136,13 +196,41 @@ class Keys(str, Enum):
...
+    return ParsedKey(key)
+
+
+def register_new_key(name: str) -> None:
+    """
+    Register a new key, e.g. "control-shift-f5", so that this can be used in a
+    key binding. (We have some validation in the key bindings that prevent the
+    creation of key bindings with unknown keys.)
+    """
+    Keys.register_new_key(name)
```
asmeurer: How do you register the corresponding ANSI code for the key?

jonathanslenders:
@asmeurer : Just a heads up that I haven't forgotten this one, just didn't had the time yet to continue working on this. I'm also looking at what it takes to support termcap/terminfo and take the escape sequences from there.

asmeurer:
Any updates on this? This is basically forcing me to stay on version 2.

> I'm also looking at what it takes to support termcap/terminfo and take the escape sequences from there.

I'm not too familiar with those. How complete are the escape sequences in those databases for things that aren't generally supported without adding custom keycodes to the terminal emulator, like Control-Shift codes?

jonathanslenders:
Hi @asmeurer: thanks for reminding me!

Some other stuff got in between. What's new however is that some time ago I merged some changes that add many control/shift/alt combinations:
- here: https://github.com/prompt-toolkit/python-prompt-toolkit/blob/master/prompt_toolkit/keys.py#L112
- and here: https://github.com/prompt-toolkit/python-prompt-toolkit/blob/master/prompt_toolkit/input/ansi_escape_sequences.py#L244

Would that already solve your issue? I certainly still want to complete this PR.

asmeurer:
@jonathanslenders the keys the I personally have added are ControlEnter, ControlSlash, ControlQuestionMark (i.e., ControlShiftSlash), EscapeUp, and EscapeDown (search for "ANSI_SEQUENCES" in https://github.com/asmeurer/mypython/blob/master/mypython/keys.py, note, in some cases I've just made up a random escape code, but if you add it to prompt-toolkit, you should use the standard from https://invisible-island.net/xterm/ctlseqs/ctlseqs.html). 

In general, I still think it's perfectly valid to define a custom escape sequence, and there should be a blessed way to do this.

I'm still pinning all of my environments to prompt-toolkit 2 just because of this issue. Many of the keys I redefined are fundamental to my workflow, so living without them is not an option.

jonathanslenders:
Ok, I have to get back to this. Thanks again @asmeurer, for the reminder.

I'm convinced now that keys should not be an enum, but can be a plain string. This is the most flexible and allows adding new keys. Also the ``ParsedKey`` that I'm using here in this PR is not needed, and it should not be required to "register" a key. However, the vt100 parser should know the mapping between the ANSI sequence and the name of the key. Ideally, this should not involve changing a global state. 

@asmeurer : Are you willing to work on this and create a PR? I guess the most important is to make sure that functions which accept a ``Key`` now accept ``str`` instead, and ``Tuple[Key, ...]`` becomes ``Tuple[str, ...]``, etc... everything else should be adjusted from there, but Mypy will probably point out what needs to change.

asmeurer:
I'm not really clear what exactly you had in mind. It seems that the Enum entries *are* strings, so the functions that accept them already accept strings. 

I was able to get this working with this patch

```patch
diff --git a/prompt_toolkit/key_binding/key_bindings.py b/prompt_toolkit/key_binding/key_bindings.py
index 06ca376b..90576ee5 100644
--- a/prompt_toolkit/key_binding/key_bindings.py
+++ b/prompt_toolkit/key_binding/key_bindings.py
@@ -52,7 +52,7 @@ from typing import (

 from prompt_toolkit.cache import SimpleCache
 from prompt_toolkit.filters import FilterOrBool, Never, to_filter
-from prompt_toolkit.keys import KEY_ALIASES, Keys
+from prompt_toolkit.keys import ALL_KEYS, KEY_ALIASES, Keys

 if TYPE_CHECKING:
     # Avoid circular imports.
@@ -447,6 +447,9 @@ def _parse_key(key: Union[Keys, str]) -> Union[str, Keys]:
     except ValueError:
         pass

+    if key in ALL_KEYS:
+        return key
+
     # Final validation.
     if len(key) != 1:
         raise ValueError("Invalid key: %s" % (key,))
diff --git a/prompt_toolkit/key_binding/key_processor.py b/prompt_toolkit/key_binding/key_processor.py
index 5d97c9ac..115a7242 100644
--- a/prompt_toolkit/key_binding/key_processor.py
+++ b/prompt_toolkit/key_binding/key_processor.py
@@ -38,8 +38,6 @@ class KeyPress:
     """

     def __init__(self, key: Union[Keys, str], data: Optional[str] = None) -> None:
-        assert isinstance(key, Keys) or len(key) == 1
-
         if data is None:
             if isinstance(key, Keys):
                 data = key.value
```

And in my application code, something like

```py
ShiftEnter = "<Shift-Enter>"
ALL_KEYS.append('<Shift-Enter>')
ANSI_SEQUENCES['\x1bOM'] = ShiftEnter
@r.add_binding(ShiftEnter)
def _(event):
    ...
```

Of course, I would still like to be able to mutate the Keys enum so that I can inject my keys as first-class, but it's not a huge deal. 

I haven't tested this extensively, but it seems to work.

asmeurer:
It looks like I can work around this in base prompt-toolkit by making my key a single character string. I can even add a longer version in KEY_ALIASES if I want. Since there's no limit to single character strings, this is not a huge deal, though it's obviously quite hacky.

PR #1017, opened 2019-12-02
fix auto_suggestion

vaaaaanquish:
fix issue: https://github.com/prompt-toolkit/python-prompt-toolkit/issues/976
PR for master: https://github.com/prompt-toolkit/python-prompt-toolkit/pull/977


jonathanslenders:
Thanks!

PR #1020, opened 2019-12-05
Protection from wrapped fileno's

dsblank:
This will keep prompt_toolkit from crashing. 

This does happen in the wild, and this should keep everything working.

### DIFF
--- a/prompt_toolkit/output/vt100.py
+++ b/prompt_toolkit/output/vt100.py
@@ -390,7 +390,12 @@ def _get_size(fileno: int) -> Tuple[int, int]:
     # Note: We should not pass 'True' as a fourth parameter to 'ioctl'. (True
     #       is the default.) This causes segmentation faults on some systems.
     #       See: https://github.com/jonathanslenders/python-prompt-toolkit/pull/364
-    fcntl.ioctl(fileno, termios.TIOCGWINSZ, buf)  # type: ignore
+    try:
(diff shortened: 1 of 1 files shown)


jonathanslenders:
Hi @dsblank,
Can you explain what it means if a fileno gets wrapped? I've never heard of that. Thanks.

dsblank:
Sorry, not a good description. This can happen when turning sys.stdout into a pipe using Python low-level tools, like os.dup2(). Everything works well (in ipython) until prompt_toolkit tries to get the size from the fake sys.stdout which is actually the pipe. Does that help?

dsblank:
I think the root of the problem is that sys.stderr *was* a tty at first (when get_size is defined) but then later is not. Let me address that, rather than the side-effect issue in this PR.

dsblank:
Yes, that was the issue. Closing this PR to check for tty at time of get_size in case it changes over time.

jonathanslenders:
Would it make sense instead to protect prompt_toolkit by dupping the fd from stdout and rendering the output to the new FD? That way people can replace stdout with anything else.

Just a thought, I didn't think it through, and not sure if it would conflict with other assumptions.

dsblank:
We already have a working system, it is just that the fileno changes from a tty at first, to a non tty later. The other PR fixes the issue, if asking if it is a tty every update is efficient enough.

PR #1021, opened 2019-12-06
Check to see if still a tty when we get_size

dsblank:
The sys.stdout may change over time from a tty to a non-tty. Can happen when putting a file-capture on the fileno.

### DIFF
--- a/prompt_toolkit/output/vt100.py
+++ b/prompt_toolkit/output/vt100.py
@@ -436,7 +436,6 @@ class Vt100_Output(Output):
             ColorDepth.DEPTH_24_BIT: _EscapeCodeCache(ColorDepth.DEPTH_24_BIT),
         }
 
-    @classmethod
     def from_pty(cls, stdout: TextIO, term: Optional[str] = None) -> "Vt100_Output":
         """
         Create an Output class from a pseudo terminal.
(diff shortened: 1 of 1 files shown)


jonathanslenders:
See also: https://github.com/prompt-toolkit/python-prompt-toolkit/pull/1020

dsblank:
Either fix is fine for us (this one or #1020 ). We probably won't be able to redesign our system, so, which one of these is better for all?

jonathanslenders:
Merging this. I'm going to add a comment to the code so that we don't forget the reason. (Otherwise there's a chance we revert this by accident.)

jonathanslenders:
Added a comment here: https://github.com/prompt-toolkit/python-prompt-toolkit/commit/7d2ef1998d0eb1ba268e9781be84051f35203b5d and made one change too. 
I don't know the details of you use case, but I believe the try/except here better protects against race conditions than the if-test. (And I prefer not to catch in `_get_size()` itself.)

Can you test, and see whether this is fine?
Thanks!

dsblank:
Works great! Thanks for the attention, and a solid product.

jonathanslenders:
You're welcome @dsblank!

PR #1024, opened 2019-12-09
Tell black to use 150 line-length

njalerikson:
Currently, only flake8 is aware of the 150 line-length rule:
https://github.com/prompt-toolkit/python-prompt-toolkit/blob/7d2ef1998d0eb1ba268e9781be84051f35203b5d/setup.cfg#L1-L3

Need to also specify in pyproject.toml for black to know about it (or at least for Sublime Text's black plugin to know about it).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.black]
 target-version = ['py36']
-
+line-length = 150
 
 [tool.isort]
 # isort configuration that is compatible with Black.


njalerikson:
Hm, I see that flake8's line length 150 is a relatively old contribution compared to using black.

So maybe the change should actually be to use line length 83 for flake8? Black's standard?

jonathanslenders:
Yes, it should be the other way around. Actually I don't want to change anything to black's defaults.
I think I kept the length for flake8 this way, so that it wouldn't complain about black's slightly longer lines in some situations. In fact, I don't care about flake8 checking line lengths, if the code is Black compliant.

PR #1025, opened 2019-12-10
[wip] ProgressBar with callable remove_when_done

njalerikson:
Driven by the conversation in #974 

Working to add the ability to provide a callable for `remove_when_done` which is run in the background and once complete, if the return value is `True`, then the counter in question is deleted.

Example use case:

```python
from prompt_toolkit.shortcuts.progress_bar import ProgressBar
import time
import random


def remove_when_done(counter):
    if counter.total % 2:
        time.sleep(2.5)
        return True


with ProgressBar() as group:
    for label in "ABCDEFGH":
        for _ in group(range(random.randrange(5, 10)), label=label, remove_when_done=remove_when_done):
            time.sleep(.1)
```

@jonathanslenders I'm still trying to wrap my head around Python's async, I don't think I'm understanding tasks correctly so I haven't been able to get this concept to work yet. The above example generates the progress bars as expected and the background tasks appear to be created however the actual removals themselves are not occurring. I'm also getting a RuntimeError when the application attempts to finalize/cleanup:

```
RuntimeError: Task <Task pending name='Task-1' coro=<...> cb=[_run_until_complete_cb() at ...]> got Future <Task cancelling name='Task-12' coro=<ProgressBarCounter._remove_when_done_async() running at ...>> attached to a different loop
```

I've been testing with 3.8.

### DIFF
--- a/examples/progress-bar/a-lot-of-parallel-tasks.py
+++ b/examples/progress-bar/a-lot-of-parallel-tasks.py
@@ -16,8 +16,28 @@ def main():
         bottom_toolbar=HTML("<b>[Control-L]</b> clear  <b>[Control-C]</b> abort"),
     ) as pb:
 
+        lock = threading.Lock()
+        completed = []
+
+        # Remove a completed counter after 5 seconds but keep the last 5.
+        def remove_when_done(counter):
+            counter.label = "Completed"
+            with lock:
+                completed.append(counter)
+            time.sleep(5)
+            counter.progress_bar.invalidate()
+            while True:
+                if len(completed) < 5:
+                    pass
+                else:
+                    if completed[0] is counter:
+                        with lock:
+                            completed.pop(0)
+                        return True
+                time.sleep(1)
+
         def run_task(label, total, sleep_time):
-            for i in pb(range(total), label=label):
+            for i in pb(range(total), label=label, remove_when_done=remove_when_done):
                 time.sleep(sleep_time)
 
         threads = []
--- a/prompt_toolkit/shortcuts/progress_bar/base.py
+++ b/prompt_toolkit/shortcuts/progress_bar/base.py
@@ -59,9 +59,7 @@ from .formatters import Formatter, create_default_formatters
 try:
     import contextvars
 except ImportError:
-    from prompt_toolkit.eventloop import (
-        dummy_contextvars as contextvars,
-    )  # type: ignore
+    import prompt_toolkit.eventloop.dummy_contextvars as contextvars  # type: ignore
 
 
 __all__ = ["ProgressBar"]
@@ -253,7 +251,9 @@ class ProgressBar:
         self,
         data: Optional[Iterable[_T]] = None,
         label: AnyFormattedText = "",
-        remove_when_done: Union[Callable[[], bool], bool] = False,
+        remove_when_done: Union[
+            Callable[["ProgressBarCounter[_T]"], bool], bool
+        ] = False,
         total: Optional[int] = None,
     ) -> "ProgressBarCounter[_T]":
         """
@@ -324,7 +324,9 @@ class ProgressBarCounter(Generic[_CounterItem]):
         progress_bar: ProgressBar,
         data: Optional[Iterable[_CounterItem]] = None,
         label: AnyFormattedText = "",
-        remove_when_done: Union[Callable[[], bool], bool] = False,
+        remove_when_done: Union[
+            Callable[["ProgressBarCounter[_CounterItem]"], bool], bool
+        ] = False,
         total: Optional[int] = None,
     ) -> None:
 
@@ -364,8 +366,8 @@ class ProgressBarCounter(Generic[_CounterItem]):
         self.items_completed += 1
         self.progress_bar.invalidate()
 
-    async def _remove_when_done_async(self):
-        def run_remove_when_done_thread() -> None:
+    async def _remove_when_done_async(self) -> None:
+        def run_remove_when_done_thread() -> bool:
             return self.remove_when_done(self)
 
         if await run_in_executor_with_context(run_remove_when_done_thread):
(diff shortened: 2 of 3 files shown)


njalerikson:
Ah, so progress bars are juggling two event loops, the main loop (`_loop`) and the application loop (`_app_loop`). The `_app_loop` is only set as the event loop in the separate app thread. Since the counters exist in the main thread it's grabbing the wrong event loop.

A slight modification was introduced to `Application.create_background_task` to where the event loop to use can be passed in. **Alternatively**, I see that `Application` is storing what appears to be its event loop:

https://github.com/prompt-toolkit/python-prompt-toolkit/blob/7d2ef1998d0eb1ba268e9781be84051f35203b5d/prompt_toolkit/application/application.py#L613-L618

So should we be using `self.loop` instead? Of passing loop or using `get_event_loop`?

jonathanslenders:
In the case of the progress bars, there should be only one event loop running, which is the one in the UI thread. ``get_event_loop()`` should never be called in the main thread. If that happens, then that's a bug.
I'll have a look.

jonathanslenders:
Two things:
- I don't really understand why ``remove_when_done`` would have to run in an executor? It should not be hard to compute, so it can run in the event loop.
- I think this can be a ``prompt_toolkit.filters.FilterOrBool``. (Use ``to_filter`` or ``is_true``, like in other places.)

njalerikson:
This allows for more creative removals. I'm personally interested in removal policies like keeping the last X counters and keeping each counter for Y minutes.

njalerikson:
Pinging this - just don't want to lose steam on this one. @jonathanslenders, let me know if there is anything I can change or further clarify.

prompt_toolkit/shortcuts/progress_bar/base.py
```diff
@@ -370,8 +382,15 @@ def done(self, value: bool) -> None:
         # If done then store the stop_time, otherwise clear.
         self.stop_time = datetime.datetime.now() if value else None
 
-        if value and self.remove_when_done:
-            self.progress_bar.counters.remove(self)
+        if value:
+            if callable(self.remove_when_done):
+                # Register a background task to run the remove_when_done callable.
```
jonathanslenders: This can be written as:
```python
self.progress_bar._app_loop.call_soon_threadsafe(
    lambda: self.progress_bar.app.create_background_task(self._remove_when_done_async()))
````

I prefer not to pass a loop argument to ``create_background_task``, because it's not thread safe and should not be called from a different thread to begin with. The only reason we have that function is to spawn coroutines in the current loop, and make sure they are cancelled when the application terminates.

File: prompt_toolkit/shortcuts/progress_bar/base.py [hunk 2]
```diff
@@ -359,6 +364,13 @@ def item_completed(self) -> None:
         self.items_completed += 1
         self.progress_bar.invalidate()
 
+    async def _remove_when_done_async(self):
+        def run_remove_when_done_thread() -> None:
+            return self.remove_when_done(self)
```
jonathanslenders: This doesn't correspond with the type signature, I think it's fine to pass ``self``, but the ``Callable`` in ``__init__`` should match.

prompt_toolkit/shortcuts/progress_bar/base.py
```diff
@@ -359,6 +364,13 @@ def item_completed(self) -> None:
         self.items_completed += 1
         self.progress_bar.invalidate()
 
+    async def _remove_when_done_async(self):
```
jonathanslenders: Add ``-> None:``

prompt_toolkit/shortcuts/progress_bar/base.py
```diff
@@ -359,6 +364,13 @@ def item_completed(self) -> None:
...
+        def run_remove_when_done_thread() -> None:
+            return self.remove_when_done(self)
+
+        if await run_in_executor_with_context(run_remove_when_done_thread):
```
jonathanslenders: I'm still thinking about whether this is the right thing to do.

By default, I'd like as much as possible for prompt_toolkit to have asynchronous APIs, and possibly create an synchronous abstraction on top of that. If we expect a blocking synchronous function here, then we exclude people that would like to include it in an async app.

njalerikson: I borrowed almost all of this code from how validation and `ThreadedValidators` are implemented. I do see that the current implementation appears to have some issue with cleaning up tasks quickly upon completion (you'll see this if you run the a-lot-of-parallel-tasks.py example that I updated).

Would it be better to allow `remove_when_done` to be either synchronous or asynchronous and then it's up to the user to design any expensive async functions correctly (with `asyncio.sleep`, etc.)?

prompt_toolkit/shortcuts/progress_bar/base.py
njalerikson: ah ok will do

prompt_toolkit/shortcuts/progress_bar/base.py
njalerikson: hm I think I got this figured out

prompt_toolkit/shortcuts/progress_bar/base.py
njalerikson: will do

examples/progress-bar/a-lot-of-parallel-tasks.py
```diff
@@ -16,8 +16,28 @@ def main():
         bottom_toolbar=HTML("<b>[Control-L]</b> clear  <b>[Control-C]</b> abort"),
     ) as pb:
 
+        lock = threading.Lock()
+        completed = []
+
+        # Remove a completed counter after 5 seconds but keep the last 5.
```
jonathanslenders: This is nice. That's a great use case.

I'm still not convinced though that a thread is the right approach here. We will get many threads that never finish and can't be joined anywhere. I'm thinking about whether we could reuse the prompt_toolkit filters in this case too, and do something like:
```python
@Condition
def ten_seconds_passed() -> bool:
    counter = get_counter()  # context local
    return time.time() - counter.stop_time > 10

@Condition
def at_least_five_bars() -> bool:
    pb = get_progress_bar()  # context local
    return len(pb.counters) > 10

for i in pb(range(total), label=label, remove_when_done=ten_seconds_passed & at_least_five_bars):
```

If this condition evaluates  during every render, we can remove the bar when this turns False.

njalerikson:
closing this in favor of #1042

PR #1029, opened 2019-12-13
Introducing ProgressBarCounter.stopped

njalerikson:
Include a distinction for counters that have stopped (time_elapsed no longer increments and Bar formatter stops running in the case of unknown totals). This allows bars to better handle cases where a counter doesn't complete properly and terminates in some error state the developer cares about.

@jonathanslenders I think this is one of the last ProgressBar ideas I have.

### DIFF
--- a/prompt_toolkit/shortcuts/progress_bar/base.py
+++ b/prompt_toolkit/shortcuts/progress_bar/base.py
@@ -56,12 +56,12 @@ from .formatters import Formatter, create_default_formatters
 try:
     import contextvars
 except ImportError:
-    from prompt_toolkit.eventloop import dummy_contextvars as contextvars  # type: ignore
+    from prompt_toolkit.eventloop import (  # type: ignore
+        dummy_contextvars as contextvars,
+    )
 
 
-__all__ = [
-    "ProgressBar",
-]
+__all__ = ["ProgressBar"]
 
 E = KeyPressEvent
 
@@ -342,13 +342,20 @@ class ProgressBarCounter(Generic[_CounterItem]):
             self.total = total
 
     def __iter__(self) -> Iterable[_CounterItem]:
-        try:
-            if self.data is not None:
+        if self.data is not None:
+            try:
                 for item in self.data:
                     yield item
                     self.item_completed()
-        finally:
-            self.done = True
+
+                # Only done if we iterate to the very end.
+                self.done = True
+            finally:
+                # Ensure counter has stopped even if we did not iterate to the
+                # end (e.g. break or exceptions).
+                self.stopped = True
+        else:
+            raise NotImplementedError("No data defined to iterate over.")
 
     def item_completed(self) -> None:
         """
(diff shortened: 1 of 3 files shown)


prompt_toolkit/shortcuts/progress_bar/base.py
```diff
@@ -361,18 +364,49 @@ def item_completed(self) -> None:
...
-
-        # If done then store the stop_time, otherwise clear.
-        self.stop_time = datetime.datetime.now() if value else None
+        self.stop = value
 
         if value and self.remove_when_done:
             self.progress_bar.counters.remove(self)
 
+    @property
+    def stop(self) -> bool:
```
jonathanslenders: Should we call it ``stopped``, because it's more like a state, so a noun instead of a verb, like ``done``.

prompt_toolkit/shortcuts/progress_bar/base.py
```diff
@@ -361,18 +364,49 @@ def item_completed(self) -> None:
...
+        (see remove_when_done) this can help aggregate failures from a large number of
+        successes.
+
+        Contrast this with done. A done counter has reached its 100% completion.
+        A stopped counter may be terminated before 100% completion.
+        """
+        return self._stop
+
+    @stop.setter
+    def stop(self, value):
```
jonathanslenders: Can you add type annotations here?

prompt_toolkit/shortcuts/progress_bar/base.py
```diff
@@ -361,18 +364,49 @@ def item_completed(self) -> None:
...
+        A stopped counter (but not done) can be used to signal that a given counter has
+        encountered an error but allows other counters to continue
+        (e.g. download X of Y failed). Given how only done counters are removed
+        (see remove_when_done) this can help aggregate failures from a large number of
+        successes.
+
+        Contrast this with done. A done counter has reached its 100% completion.
+        A stopped counter may be terminated before 100% completion.
+        """
+        return self._stop
```
jonathanslenders: I think we can remove the ``self._stop`` variable and do ``return self.stop_time is not None`` here.

jonathanslenders:
This makes sense, I've added a few remarks. Thanks a lot for this!

prompt_toolkit/shortcuts/progress_bar/base.py
njalerikson: No problem

prompt_toolkit/shortcuts/progress_bar/base.py
njalerikson: Will do

prompt_toolkit/shortcuts/progress_bar/base.py
njalerikson: Agreed

prompt_toolkit/shortcuts/progress_bar/base.py
```diff
@@ -342,13 +342,15 @@ def __init__(
...
-                for item in self.data:
-                    yield item
-                    self.item_completed()
-        finally:
+        if self.data is not None:
+            for item in self.data:
+                yield item
+                self.item_completed()
+
+            # Only done if we iterate to the very end.
```
njalerikson: should we intercept any exceptions here and set stopped?

jonathanslenders:
I'm merging this. It looks good. We can still improve/fix in other PRs.

Wouldn't it make sense to set the ETA to 0 when stopped=True? This confused me in the example.

PR #1038, opened 2019-12-20
Set time left to 0 when done or stopped

njalerikson:
From https://github.com/prompt-toolkit/python-prompt-toolkit/pull/1029#issuecomment-568082229

When a counter is `done` or `stopped` zero the time left irrelevant of the percentage completed.


jonathanslenders:
Thank you @njalerikson!

PR #1042, opened 2019-12-23
[wip] Initial attempt for a Conditional remove_when_done

njalerikson:
Inspired by the discussion in https://github.com/prompt-toolkit/python-prompt-toolkit/pull/1025#discussion_r360580489.

Instead of using a threaded approach for `remove_when_done` processing we check for removals as part of the rendering process instead.

@jonathanslenders this feels like a hack at this point, not sure if we'd want the `get_progress_bar`/`set_progress_bar`/`get_counter`/`set_counter` to be extending the `AppSession` or whether using this separate `ProgressBarSession` is preferred.

If we continue down this contexvar path then maybe `AppSession` should be made more generic to where anything can be set in a session? So we'd introduce something like `get_contextvar(key)` and `set_contextvar(key, value)`. Just thinking out loud here.


njalerikson:
Pinging this.

Been using this change for the last week and haven't run into any issues. So I don't have any concerns with implementing `remove_when_done` as a `Conditional`.

The only open question is whether the `get_progress_bar` and `get_counter` are implemented in the desired way.

njalerikson:
@jonathanslenders let me know if this is of interest or whether I should just close this and implement a monkey patch on my end - thanks

PR #1043, opened 2019-12-26
Fix the style name: current-line-number -> line-number.current

hokuda:


jonathanslenders:
Thank you! That is a bug indeed.

I believe ``class:line-number.current`` automatically expands to ``class:line-number,line-number.current``. See: https://python-prompt-toolkit.readthedocs.io/en/master/pages/advanced_topics/styling.html#dot-notation-in-class-names

So, no need to mention both. Can you fix that (and squash in the same commit)? Then I'll merge it.

jonathanslenders:
Thank you @hokuda!

PR #1047, opened 2020-01-07
Improved the performance of the regular languages compiler: generate …

jonathanslenders:
…fewer and better regexes.

### DIFF
--- a/prompt_toolkit/contrib/regular_languages/__init__.py
+++ b/prompt_toolkit/contrib/regular_languages/__init__.py
@@ -54,17 +54,16 @@ named group will test the captured string.
 
 There is one tricky bit:
 
-    Ofter we operate on incomplete input (this is by definition the case for
+    Often we operate on incomplete input (this is by definition the case for
     autocompletion) and we have to decide for the cursor position in which
     possible state the grammar it could be and in which way variables could be
     matched up to that point.
 
 To solve this problem, the compiler takes the original regular expression and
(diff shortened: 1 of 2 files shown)


laixintao:
Hi @jonathanslenders thanks for the work. The compiling time saved a lot.

I read the code and think it's good, but there is a problem with completion on space.

eg:
On this branch, when I type `SPACE` the completion didn't show up:
Screen Shot 2020-01-08 at 10 34 43 AM

Screen Shot 2020-01-08 at 10 34 46 AM

But on prompt_toolkit(2.0 which is iredis current use) it works fine:

Screen Shot 2020-01-08 at 10 39 13 AM

jonathanslenders:
@laixintao : Thank you for the feedback! I really appreciate you trying it out.
This bug needs to be fixed before merging it. I'll look into that.

jonathanslenders:
@laixintao: can you check again against this branch?
I did a few bug fixes, and maybe the performance is a bit better even.

I *think* I got it right, and the above bug is solved.
Still need to clean up the code, before merging.

laixintao:
@jonathanslenders Thanks!

I tested with this branch, the cost time down to `.7 seconds, and all the compilations work fine as far as I tested.

From my log, the _re_prefix_patterns that need to compile down to 496 only (previous ~8000), that's amazing! Excellent work!

```
 INFO Compiled _re_prefix_pattern: 496
DEBUG [compile] Compile finished! Cost: 1.7887732982635498
```

PR #1064, opened 2020-01-26
set_exeption_handler improvements

Chris3606:
This is just a couple minor tweaks.
- `Application.run` was taking a `set_exception_handler` parameter but never using it, so I just added the code to forward it to the `Application.run_async` call. 
- Added `set_exception_handler` parameters to `PromptSession`.  I've found it's useful even in the context of a shortcut class like `PromptSession` to replace the application's default exception handler if you, for example, want to do something like exit the application on exception.


jonathanslenders:
Thank you for this, @Chris3606!

PR #1069, opened 2020-01-30
Name all underscore functions. -> Make Mypy happy.

jonathanslenders:


asmeurer:
+1. This also makes pyflakes happy, and it can be useful to have these functions be reusable.

PR #1074, opened 2020-02-09
bugfix: nested completion with leading space

crhan:
NestedCompleter meets a bug when input document has some leading space


jonathanslenders:
Thank you both!

PR #1096, opened 2020-03-11
Drop last use of six

chronitis:
This appears to be the only remaining reference to `six`, which can probably be dropped completely as a test requirement now that python 2 compat is no longer required.


jonathanslenders:
Thank you @chronitis !

PR #1112, opened 2020-04-01
Added fancy-zsh-prompt.py example.

jonathanslenders:


examples/prompts/fancy-zsh-prompt.py
```diff
@@ -0,0 +1,72 @@
+#!/usr/bin/env python
+"""
+Example of the fancy ZSH prompt that Chris Lasher was using (see issue #1111).
```
gotgenes, line 3: Original source is from @anki-code in https://github.com/xonsh/xonsh/issues/3356#issue-506263409

PR #1116, opened 2020-04-09
Ensure rprompt aligns with input.

fshowalter:
Align the rprompt with the input line in the case of a multi-line prompt text. Since this is a change on the float, we can't simply style the rprompt to get the desired intent.


jonathanslenders:
I did not yet merge/respond to this, because I've been wondering whether this would break other people's expectations. I'm not sure though. I'll merge it for now, but revert the change if it turns out to be an issue. In that case we should make it configurable.

PR #1118, opened 2020-04-11
Updated documentation for pattern parameter of WordCompleter

voidcase:
I found myself confused about the `pattern` parameter. Specifically what is it used for, and what is this "default one" that it replaces? Updated the docstring to clarify this.


jonathanslenders:
Thanks!

Issue #1139, opened 2020-05-09
Getting Unhandled exception in event loop: Press Enter to continue


yerfdogyrag:
This appears to be due to cpr events getting lost along the way.  This little bit of code reproduces the issue under linux (python 3.7.3, prompt_toolkit 3.0.5):

```
import os
from asyncio import sleep, get_running_loop, run
from prompt_toolkit.patch_stdout import patch_stdout
from prompt_toolkit.shortcuts import PromptSession

from prompt_toolkit import __version__
print("Version:", __version__)


def _exception_handler(loop, context):
    # Should never get here - exit quickly with error
    os._exit(4)


async def t1():
    'Print stuff to screen quickly so prompt_toolkit needs cpr'
    counter = 0 
    while True:
        print("Counter:", counter)
        counter += 1 
        await sleep(.01)


async def main():
    loop = get_running_loop()
    loop.set_exception_handler(_exception_handler)
    loop.create_task(t1())
    session = PromptSession('% ')
    with patch_stdout():
        await session.prompt_async(set_exception_handler=False)

run(main())
```

This appears to be due to a cpr event getting lost along the way and then the wait_for_responses task eventually gets reclaimed and we get the error.  I'll submit a pr for discussion.

PR #1140, opened 2020-05-09
Fix issue with asyncio task getting lost during text storm.  #1139

yerfdogyrag:
I believe this will need to be fixed in python 3.8 anyway - they're deprecating having wait actually create tasks.  I'd prefer using create_task (with a name!) here, but that would probably break 3.6.

Thoughts?

### DIFF
--- a/prompt_toolkit/renderer.py
+++ b/prompt_toolkit/renderer.py
@@ -2,7 +2,7 @@
 Renders the command line on the console.
 (Redraws parts of the input line that were changed.)
 """
-from asyncio import FIRST_COMPLETED, Future, sleep, wait, ensure_future, CancelledError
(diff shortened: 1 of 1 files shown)


jonathanslenders:
Looks like there was some conflict with this PR: https://github.com/prompt-toolkit/python-prompt-toolkit/pull/1141 that tries to solve somewhat the same thing.

But I guess if we cancel all the futures like we do now, we don't have to use "wait". we can do ``await wait_for_responses()`` and then cancel the ``wait_for_timeout`` task if that's still ongoing.

yerfdogyrag:
#1141 seems to fix the same issue that I had.   I think the only reason to make a change now is due to cleanliness and the wait/coroutine usage being deprecated in 3.8.  It's up to you - I can combine #1141 and do the await the way you mention above or just drop the PR.

Thanks!

Issue #876, opened 2019-04-13
Posix pipe inputs might respond to CPR


vxgmichel:
*Issue # 5 from [this gist](https://gist.github.com/vxgmichel/f99c372f05dafb16ab24960d985ba61a)*

It'd be nice to be able to configure whether a posix pipe input responds to CPR. It is currently forced to `False`:

https://github.com/prompt-toolkit/python-prompt-toolkit/blob/9e3845abe9d3cd64c096eaba8d890943f86b8954/prompt_toolkit/input/posix_pipe.py#L39-L41

But this pipe might be connected to a vt100 terminal that supports ANSI codes. A good example of that is `TelnetConnection.vt100_input`:

https://github.com/prompt-toolkit/python-prompt-toolkit/blob/9e3845abe9d3cd64c096eaba8d890943f86b8954/prompt_toolkit/contrib/telnet/server.py#L112

In this context `telnet_connection.vt100_input.responds_to_cpr` returns `False` even though CPR is most likely supported.

Also, all the tests seems to pass with `PosixPipeInput.responds_to_cpr` forced to `True` instead.

jonathanslenders:
@vxgmichel: thanks for pointing out! I think you're right and this should be fixed. I think we can take a ``responds_to_cpr`` boolean input in ``PosixPipeInput`` that defaults to False.

PR #1150, opened 2020-05-21
Better support for remote terminals

vxgmichel:
(Fix https://github.com/prompt-toolkit/python-prompt-toolkit/issues/876)

### Context

I'm trying to serve a ptpython console using both telnet and ssh:
https://gist.github.com/vxgmichel/7685685b3e5ead04ada4a3ba75a48eef

With the recent efforts to port prompt-toolkit to asyncio, it feels like the project is really close to have a nice and clean separation between the terminal on one side (thanks to `create_app_session`) and the application on the other side. Notice how simple the gist is, especially compared to [the one I wrote last year](https://gist.github.com/vxgmichel/f99c372f05dafb16ab24960d985ba61a) for prompt-toolkit 2.

However, I ran into a few problems when trying to run this example on linux and windows (with telnet and ssh clients running on both linux and windows). Some were related to ptpython itself and I'll make a separate PR on the ptpython repo.

### Issues

On the prompt-toolkit side, I noticed the following issues:

- Pipe input objects do not respond to CPR even though the client most probably does (see #876). I made `responds_to_cpr` configurable and defaulting to `True` for pipe input objects since it is assumed that they support `VT100`.

- Some code, mostly in `renderer.py`, uses `is_windows()` to perform some specific operations for local windows terminal. The problem is that one might use windows to serve an application to remote clients. The commit cb901ba implements `output.is_windows()` for this purpose. It also has to tweak a bit how the default color depth is chosen, for the same reasons.

- The telnet and ssh servers were not compatible with windows, but more importantly they did not provide the client terminal type to the prompt-toolkit session. Both protocol support having the client report its terminal type, so getting and setting this type as `output.term` opens the way to better remote terminal support (typically using the right color depth for the client terminal).

- The ssh server replaces LF by CRLF before writing to the client stream, but the telnet client does not. I fixed it in the telnet server for consistency even though I'm a bit puzzled by the CRLF handling in general. From what I can tell, `print_formatted_text` [does perform this replacing already](https://github.com/prompt-toolkit/python-prompt-toolkit/blob/a49c486fa5ae6ad5a38ff3c1050f9c9609acac1d/prompt_toolkit/renderer.py#L769), but it also happens that the `output.write()` method is used directly in some case. I could only find [one occurrence](https://github.com/prompt-toolkit/python-prompt-toolkit/blob/a49c486fa5ae6ad5a38ff3c1050f9c9609acac1d/prompt_toolkit/shortcuts/prompt.py#L1037)) in `prompt-toolkit` and it does use CRLF. However, there is plenty of write calls in the ptpython REPL that uses LF instead (e.g [when rendering an exception](https://github.com/prompt-toolkit/ptpython/blob/89017ba158ed1d95319233fa5aedf3931c3b8b77/ptpython/repl.py#L241)).

### Proof of concept

This PR is a proof of concept and is probably not ready to merge as it is, especially since it changes some part of the prompt-toolkit API (e.g `ColorDepth.default`).

### DIFF
--- a/prompt_toolkit/contrib/telnet/server.py
+++ b/prompt_toolkit/contrib/telnet/server.py
@@ -81,6 +81,7 @@ class _ConnectionStdout:
         self._buffer: List[bytes] = []
 
     def write(self, data: str) -> None:
+        data = data.replace("\n", "\r\n")
         self._buffer.append(data.encode(self._encoding, errors=self._errors))
         self.flush()
 
--- a/prompt_toolkit/contrib/telnet/server.py
+++ b/prompt_toolkit/contrib/telnet/server.py
@@ -11,7 +11,7 @@ from prompt_toolkit.application.current import create_app_session, get_app
 from prompt_toolkit.application.run_in_terminal import run_in_terminal
 from prompt_toolkit.data_structures import Size
 from prompt_toolkit.formatted_text import AnyFormattedText, to_formatted_text
-from prompt_toolkit.input.posix_pipe import PosixPipeInput
+from prompt_toolkit.input import create_pipe_input
 from prompt_toolkit.output.vt100 import Vt100_Output
 from prompt_toolkit.renderer import print_formatted_text as print_formatted_text
 from prompt_toolkit.styles import BaseStyle, DummyStyle
@@ -26,7 +26,9 @@ from .protocol import (
     NAWS,
     SB,
     SE,
+    SEND,
     SUPPRESS_GO_AHEAD,
+    TTYPE,
     WILL,
     TelnetProtocolParser,
 )
@@ -58,6 +60,12 @@ def _initialize_telnet(connection: socket.socket) -> None:
 
     # Negotiate window size
     connection.send(IAC + DO + NAWS)
+
+    # Negotiate terminal type
+    connection.send(IAC + DO + TTYPE)
+
+    # Negotiate terminal type
+    connection.send(IAC + SB + TTYPE + SEND + IAC + SE)
 
 
 class _ConnectionStdout:
@@ -115,6 +123,8 @@ class TelnetConnection:
         self.encoding = encoding
         self.style = style
         self._closed = False
+        self._ready = asyncio.Event()
+        self.vt100_output = None
 
         # Create "Output" object.
         self.size = Size(rows=40, columns=79)
@@ -123,14 +133,13 @@ class TelnetConnection:
         _initialize_telnet(conn)
 
         # Create input.
-        self.vt100_input = PosixPipeInput()
+        self.vt100_input = create_pipe_input()
 
         # Create output.
         def get_size() -> Size:
             return self.size
 
         self.stdout = cast(TextIO, _ConnectionStdout(conn, encoding=encoding))
-        self.vt100_output = Vt100_Output(self.stdout, get_size, write_binary=False)
 
         def data_received(data: bytes) -> None:
             """ TelnetProtocolParser 'data_received' callback """
@@ -139,9 +148,17 @@ class TelnetConnection:
         def size_received(rows: int, columns: int) -> None:
             """ TelnetProtocolParser 'size_received' callback """
             self.size = Size(rows=rows, columns=columns)
-            get_app()._on_resize()
+            if self.vt100_output is not None:
+                get_app()._on_resize()
 
-        self.parser = TelnetProtocolParser(data_received, size_received)
+        def ttype_received(ttype: str) -> None:
+            """ TelnetProtocolParser 'ttype_received' callback """
+            self.vt100_output = Vt100_Output(
+                self.stdout, get_size, term=ttype, write_binary=False
+            )
+            self._ready.set()
+
+        self.parser = TelnetProtocolParser(data_received, size_received, ttype_received)
         self.context: Optional[contextvars.Context] = None
 
     async def run_application(self) -> None:
@@ -158,25 +175,24 @@ class TelnetConnection:
                 logger.info("Connection closed by client. %r %r" % self.addr)
                 self.close()
 
-        async def run() -> None:
-            # Add reader.
-            loop = get_event_loop()
-            loop.add_reader(self.conn, handle_incoming_data)
+        # Add reader.
+        loop = get_event_loop()
+        loop.add_reader(self.conn, handle_incoming_data)
 
-            try:
+        try:
+            # Wait for v100_output to be properly instantiated
+            await self._ready.wait()
+            with create_app_session(input=self.vt100_input, output=self.vt100_output):
+                self.context = contextvars.copy_context()
                 await self.interact(self)
-            except Exception as e:
-                print("Got %s" % type(e).__name__, e)
-                import traceback
+        except Exception as e:
+            print("Got %s" % type(e).__name__, e)
+            import traceback
 
-                traceback.print_exc()
-                raise
-            finally:
-                self.close()
-
-        with create_app_session(input=self.vt100_input, output=self.vt100_output):
-            self.context = contextvars.copy_context()
-            await run()
+            traceback.print_exc()
+            raise
+        finally:
+            self.close()
 
     def feed(self, data: bytes) -> None:
         """
@@ -199,6 +215,8 @@ class TelnetConnection:
         """
         Send text to the client.
         """
+        if self.vt100_output is None:
+            return
         formatted_text = to_formatted_text(formatted_text)
         print_formatted_text(
             self.vt100_output, formatted_text, self.style or DummyStyle()
@@ -224,6 +242,8 @@ class TelnetConnection:
         """
         Erase the screen and move the cursor to the top.
         """
+        if self.vt100_output is None:
+            return
         self.vt100_output.erase_screen()
         self.vt100_output.cursor_goto(0, 0)
         self.vt100_output.flush()
--- a/prompt_toolkit/contrib/telnet/server.py
+++ b/prompt_toolkit/contrib/telnet/server.py
@@ -62,9 +62,12 @@ def _initialize_telnet(connection: socket.socket) -> None:
     connection.send(IAC + DO + NAWS)
 
     # Negotiate terminal type
+    # Assume the client will accept the negociation with `IAC +  WILL + TTYPE`
     connection.send(IAC + DO + TTYPE)
 
-    # Negotiate terminal type
+    # We can then select the first terminal type supported by the client,
+    # which is generally the best type the client supports
+    # The client should reply with a `IAC + SB  + TTYPE + IS + ttype + IAC + SE`
     connection.send(IAC + SB + TTYPE + SEND + IAC + SE)
 
 
--- a/prompt_toolkit/output/color_depth.py
+++ b/prompt_toolkit/output/color_depth.py
@@ -35,7 +35,7 @@ class ColorDepth(str, Enum):
     TRUE_COLOR = DEPTH_24_BIT
 
     @classmethod
-    def default(cls, term: Optional[str] = None) -> "ColorDepth":
+    def local_default(cls, term: Optional[str] = None) -> "ColorDepth":
         """
         Return the default color depth, according to the $TERM value.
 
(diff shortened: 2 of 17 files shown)


jonathanslenders:
Hi @vxgmichel,

Thanks for this! I quickly went through the changes, and in general I like the idea and how it's implemented. I still need to take some time to really go through this more in depth. Hope to do that soon.

prompt_toolkit/contrib/telnet/server.py
```diff
@@ -59,6 +61,12 @@ def _initialize_telnet(connection: socket.socket) -> None:
     # Negotiate window size
     connection.send(IAC + DO + NAWS)
 
+    # Negotiate terminal type
```
jonathanslenders, line 64: Both sequences here have the comment "Negotiate terminal type".
Are they doing the same thing?

prompt_toolkit/output/color_depth.py
```diff
@@ -62,7 +69,8 @@ def default(cls, term: Optional[str] = None) -> "ColorDepth":
         # vt100 escape sequences with ENABLE_VIRTUAL_TERMINAL_PROCESSING are
         # supported. We don't have a reliable way yet to know whether our
         # console supports true color or only 4-bit.
-        if is_windows() and "PROMPT_TOOLKIT_COLOR_DEPTH" not in os.environ:
+        windows = output.is_windows() if output is not None else is_windows()
+        if windows and "PROMPT_TOOLKIT_COLOR_DEPTH" not in os.environ:
```
jonathanslenders, line 73: I think here we probably don't need a specialization for Windows, but instead, the Output classes can have a method
``get_default_color_depth()`` which is called here.

prompt_toolkit/output/color_depth.py
```diff
@@ -46,9 +50,12 @@ def default(cls, term: Optional[str] = None) -> "ColorDepth":
         override this outcome. This is a way to enforce a certain color depth
         in all prompt_toolkit applications.
 
-        If no `term` parameter is given, we use the $TERM environment variable.
+        If no `output` parameter is given or it does not provide a corresponding
+        `term`, we use the $TERM environment variable.
         """
         # Take `TERM` value from environment variable if nothing was passed.
+        term = getattr(output, "term", None)
```
jonathanslenders, line 57: Like the comment below, this code where we check "term" can probably move to ``Vt100Output.get_default_color_depth()``.

prompt_toolkit/output/windows10.py
```diff
@@ -52,6 +52,8 @@ def flush(self) -> None:
 
     def __getattr__(self, name: str) -> Any:
         if name in (
+            "term",
```
jonathanslenders, line 55: If we have ``Output.get_default_color_depth()``, then we don't need these two.

prompt_toolkit/renderer.py
```diff
@@ -742,7 +743,7 @@ def print_formatted_text(
     """
     fragments = to_formatted_text(formatted_text)
     style_transformation = style_transformation or DummyStyleTransformation()
-    color_depth = color_depth or ColorDepth.default()
+    color_depth = color_depth or ColorDepth.default(output)
```
jonathanslenders, line 746: ``color_depth or output.get_default_color_depth()``

jonathanslenders:
Hi @vxgmichel , I added a few comments. Can you have a look at these? Maybe also rebase on the master branch, so that the false-positive mypy errors are fixed.

prompt_toolkit/contrib/telnet/server.py
vxgmichel: The full exchange for selecting the first terminal type supported by the client (which is generally the best type the client supports) is actually:
```
[Server]> IAC DO TERMINAL-TYPE
[Client]< IAC WILL TERMINAL-TYPE
[Server]> IAC SB TERMINAL-TYPE SEND IAC SE
[Client]< IAC SB TERMINAL-TYPE IS "VT100" IAC SE
```
Source: http://mud-dev.wikidot.com/telnet:negotiation

Here, I don't wait for `IAC WILL TERMINAL-TYPE` because I didn't want those changes to get too complicated as it seems safe to assume that the client will accept the terminal type negociation.

Is adding a comment enough or do you want me to implement the negotiation more rigorously?

prompt_toolkit/output/color_depth.py
vxgmichel: Right, that makes sense!

vxgmichel:
Hi @jonathanslenders 

> Can you have a look at these? Maybe also rebase on the master branch, so that the false-positive mypy errors are fixed.

Done!

I rewrote the history of the branch, here's the list of changes:
- https://github.com/prompt-toolkit/python-prompt-toolkit/pull/1150/commits/b2797d5bf4e2014e370abc3acde66216b7337d35: Pass the prompt-toolkit ssh session to the `interact` coroutine
 It's quite useful to get the ssh session object in the `interact` coroutine, plus the telnet server already does that with the `TelnetConnection` object. I had to rename `PromptToolkitSession` with `PromptToolkitSSHSession` so it's not confused with regulars prompt sessions.

- https://github.com/prompt-toolkit/python-prompt-toolkit/pull/1150/commits/b9663182e6dd5e5ef76f643cfe23285eaeb7c9c1 Add `get_default_color_depth` method to Output objects 
 I added the `get_default_color_depth()` as you asked and turned `ColorDepth.default()` into three methods:
  - `ColorDepth.local_default(term)`
  - `ColorDepth.vt100_default(term)`
  - `ColorDepth.windows_default()`

 The problem is that as far as I know, there's no mechanism in the code base to tell whether the output object corresponds to a local or remote terminal. This is important because environment variables like `TERM` or `PROMPT_TOOLKIT_COLOR_DEPTH` should be ignored for remote terminals. Do you know how you'd like to see this implemented?

- https://github.com/prompt-toolkit/python-prompt-toolkit/pull/1150/commits/f921e24bd58be3956992065b4e2a999a10993924 Remove is_windows() calls from the renderer 
  Here, checking for existing methods looked fine to me but let me know if you want to have another approach.

- https://github.com/prompt-toolkit/python-prompt-toolkit/pull/1150/commits/cd4e4c3e9b3bd9eb4f867789112428e95ee0ba0e Add a comment explaining terminal type negociation in telnet 
 See this comment https://github.com/prompt-toolkit/python-prompt-toolkit/pull/1150#discussion_r441450772

vxgmichel:
@jonathanslenders I made a few mistakes in my commits and had to force-push a couple of times. The branch should be ready to review now :)

prompt_toolkit/contrib/telnet/server.py
jonathanslenders: A comment is enough. We can always improve the implementation if needed for somebody.

prompt_toolkit/output/vt100.py
```diff
@@ -683,3 +683,6 @@ def bell(self) -> None:
         " Sound bell. "
         self.write_raw("\a")
         self.flush()
+
+    def get_default_color_depth(self) -> ColorDepth:
+        return ColorDepth.vt100_default(self.term)
```
jonathanslenders, line 688: I would actually move the implementation of ``vt100_default`` in here. (and same for the other Output implementations.)

Basically the open/closed principle: no need to change the ``ColorDepth`` class if we define a new ``Output``.
I can make that change.

jonathanslenders:
> The problem is that as far as I know, there's no mechanism in the code base to tell whether the output object corresponds to a local or remote terminal. This is important because environment variables like TERM or PROMPT_TOOLKIT_COLOR_DEPTH should be ignored for remote terminals. Do you know how you'd like to see this implemented?

I think about that. thanks a lot for all the work you've put in this PR so far.

vxgmichel:
@jonathanslenders 

Ahah I didn't expect you to merge this PR so soon! 

I just want to make it clear that **this PR broke the use of local terminal configuration**  (i.e the use of `TERM` and `PROMPT_TOOLKIT_COLOR_DEPTH` environment variables for determining color depth)

Here's a possible fix:

```diff
diff --git a/prompt_toolkit/contrib/ssh/server.py b/prompt_toolkit/contrib/ssh/server.py
index c495b04a..0e114246 100644
--- a/prompt_toolkit/contrib/ssh/server.py
+++ b/prompt_toolkit/contrib/ssh/server.py
@@ -71,7 +71,11 @@ class PromptToolkitSSHSession(asyncssh.SSHServerSession):
         term = self._chan.get_terminal_type()
 
         self._output = Vt100_Output(
-            self.stdout, self._get_size, term=term, write_binary=False
+            self.stdout,
+            self._get_size,
+            term=term,
+            write_binary=False,
+            use_local_term_configuration=False,
         )
         with create_app_session(input=self._input, output=self._output) as session:
             self.app_session = session
diff --git a/prompt_toolkit/contrib/telnet/server.py b/prompt_toolkit/contrib/telnet/server.py
index 5a6b1271..b13d2c21 100644
--- a/prompt_toolkit/contrib/telnet/server.py
+++ b/prompt_toolkit/contrib/telnet/server.py
@@ -158,7 +158,11 @@ class TelnetConnection:
         def ttype_received(ttype: str) -> None:
             """ TelnetProtocolParser 'ttype_received' callback """
             self.vt100_output = Vt100_Output(
-                self.stdout, get_size, term=ttype, write_binary=False
+                self.stdout,
+                get_size,
+                term=ttype,
+                write_binary=False,
+                use_local_term_configuration=False,
             )
             self._ready.set()
 
diff --git a/prompt_toolkit/output/vt100.py b/prompt_toolkit/output/vt100.py
index b71a3c2f..6663d973 100644
--- a/prompt_toolkit/output/vt100.py
+++ b/prompt_toolkit/output/vt100.py
@@ -403,6 +403,9 @@ class Vt100_Output(Output):
     :param term: The terminal environment variable. (xterm, xterm-256color, linux, ...)
     :param write_binary: Encode the output before writing it. If `True` (the
         default), the `stdout` object is supposed to expose an `encoding` attribute.
+    :param use_local_term_configuration: Use the local terminal configuration
+        (i.e`TERM` and `PROMPT_TOOLKIT_COLOR_DEPTH` environment variables)
+        to determine the color depth.
     """
 
     # For the error messages. Only display "Output is not a terminal" once per
@@ -415,6 +418,7 @@ class Vt100_Output(Output):
         get_size: Callable[[], Size],
         term: Optional[str] = None,
         write_binary: bool = True,
+        use_local_term_configuration: bool = True,
     ) -> None:
 
         assert all(hasattr(stdout, a) for a in ("write", "flush"))
@@ -427,6 +431,7 @@ class Vt100_Output(Output):
         self.write_binary = write_binary
         self._get_size = get_size
         self.term = term
+        self.use_local_term_configuration = use_local_term_configuration
 
         # Cache for escape codes.
         self._escape_code_caches: Dict[ColorDepth, _EscapeCodeCache] = {
@@ -685,4 +690,6 @@ class Vt100_Output(Output):
         self.flush()
 
     def get_default_color_depth(self) -> ColorDepth:
+        if self.use_local_term_configuration:
+            return ColorDepth.local_default(self.term)
         return ColorDepth.vt100_default(self.term)
diff --git a/prompt_toolkit/output/win32.py b/prompt_toolkit/output/win32.py
index 887c1d40..88d6166d 100644
--- a/prompt_toolkit/output/win32.py
+++ b/prompt_toolkit/output/win32.py
@@ -83,10 +83,20 @@ class Win32Output(Output):
     """
     I/O abstraction for rendering to Windows consoles.
     (cmd.exe and similar.)
+
+    :param use_local_term_configuration: Use the local terminal configuration
+        (i.e`TERM` and `PROMPT_TOOLKIT_COLOR_DEPTH` environment variables)
+        to determine the color depth.
     """
 
-    def __init__(self, stdout: TextIO, use_complete_width: bool = False) -> None:
+    def __init__(
+        self,
+        stdout: TextIO,
+        use_complete_width: bool = False,
+        use_local_term_configuration: bool = True,
+    ) -> None:
         self.use_complete_width = use_complete_width
+        self.use_local_term_configuration = use_local_term_configuration
 
         self._buffer: List[str] = []
         self.stdout = stdout
@@ -480,6 +490,8 @@ class Win32Output(Output):
         windll.user32.RedrawWindow(handle, None, None, c_uint(RDW_INVALIDATE))
 
     def get_default_color_depth(self) -> ColorDepth:
+        if self.use_local_term_configuration:
+            return ColorDepth.local_default()
         return ColorDepth.windows_default()
 
 
```

prompt_toolkit/output/vt100.py
vxgmichel: That makes sense! I tried to keep the code for choosing a color depth at the same place to make it clear which logic is used in which case.

PR #1176, opened 2020-06-29
Move colordepth logic to Output implementations and added ignore_envi…

jonathanslenders:
…ronment_variables parameter to AppSession.

### DIFF
--- a/prompt_toolkit/application/application.py
+++ b/prompt_toolkit/application/application.py
@@ -268,6 +268,7 @@ class Application(Generic[_AppResult]):
         session = get_app_session()
         self.output = output or session.output
         self.input = input or session.input
+        self.ignore_environment_variables = session.ignore_environment_variables
 
         # List of 'extra' functions to execute before a Application.run.
         self.pre_run_callables: List[Callable[[], None]] = []
@@ -364,15 +365,27 @@ class Application(Generic[_AppResult]):
     @property
     def color_depth(self) -> ColorDepth:
         """
-        Active :class:`.ColorDepth`.
+        The active :class:`.ColorDepth`.
+
+        The current value is determined as follows:
+        - If a color depth was given explicitely to this application, use that
+          value.
+        - If a color depth was specified in the $PROMPT_TOOLKIT_COLOR_DEPTH
+          environment variable, and `color_depth_from_env` is set, use this
+          color depth.
+        - Finally, fall back to the color depth that is reported by the
+          :class:`.Output` implementation.
         """
         depth = self._color_depth
 
         if callable(depth):
-            return depth() or self.output.get_default_color_depth()
+            depth = depth()
+
+        if depth is None and not self.ignore_environment_variables:
+            depth = ColorDepth.from_env()
 
         if depth is None:
-            return self.output.get_default_color_depth()
+            depth = self.output.get_default_color_depth()
 
         return depth
 
--- a/prompt_toolkit/application/current.py
+++ b/prompt_toolkit/application/current.py
@@ -37,14 +37,23 @@ class AppSession:
         running in this session, unless an input is passed to the `Application`
         explicitely.
     :param output: Use this as a default output.
+    :param ignore_environment_variables: Ignore environment
+        variables that don't make sense when running the Application over
+        an SSH/Telnet session. For instance, the $PROMPT_TOOLKIT_COLOR_DEPTH
+        environment variable is used to configure applications that are running
+        locally.
     """
 
     def __init__(
-        self, input: Optional["Input"] = None, output: Optional["Output"] = None
+        self,
+        input: Optional["Input"] = None,
+        output: Optional["Output"] = None,
+        ignore_environment_variables: bool = False,
     ) -> None:
 
         self._input = input
         self._output = output
+        self.ignore_environment_variables = ignore_environment_variables
 
         # The application will be set dynamically by the `set_app` context
         # manager. This is called in the application itself.
@@ -140,7 +149,9 @@ def set_app(app: "Application[Any]") -> Generator[None, None, None]:
 
 @contextmanager
 def create_app_session(
-    input: Optional["Input"] = None, output: Optional["Output"] = None
+    input: Optional["Input"] = None,
+    output: Optional["Output"] = None,
+    ignore_environment_variables: Optional[bool] = None,
 ) -> Generator[AppSession, None, None]:
     """
     Create a separate AppSession.
@@ -158,9 +169,15 @@ def create_app_session(
         input = get_app_session().input
     if output is None:
         output = get_app_session().output
+    if ignore_environment_variables is None:
+        ignore_environment_variables = get_app_session().ignore_environment_variables
 
     # Create new `AppSession` and activate.
-    session = AppSession(input=input, output=output)
+    session = AppSession(
+        input=input,
+        output=output,
+        ignore_environment_variables=ignore_environment_variables,
+    )
 
     token = _current_app_session.set(session)
     try:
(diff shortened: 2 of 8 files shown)


jonathanslenders:
Follow-up of https://github.com/prompt-toolkit/python-prompt-toolkit/pull/1150

@vxgmichel : What do you think? It's a bit different from your proposal.
I'm not sure about the ``ignore_environment_variables`` varname. I think it can probably be used for all os.environ dependent stuff in the end. I didn't want this code in the Output classes, because we would end up repeating that for all implementations.

I did not yet take the time to test this change extensively.

vxgmichel:
I just had a look at those changes and realized you ended up taking the `TERM` environment variables out of the internal logic. The only place it is being used now is here in `create_output`:
https://github.com/prompt-toolkit/python-prompt-toolkit/blob/6cd5310ecb480c0b1f54f75aab1cf19132268f2e/prompt_toolkit/output/defaults.py#L62

And I think this approach makes a lot of sense: instead of using the env information in the code internals and passing arguments around to enable/disable environment configuration, it is simply used as a default at the highest level if it is not already provided.

Maybe the same approach can be used for `PROMPT_TOOLKIT_COLOR_DEPTH`? Consider the following (incomplete) diff:
```diff
diff --git a/prompt_toolkit/application/application.py b/prompt_toolkit/application/application.py
index 48b4d280..04ebc5f1 100644
--- a/prompt_toolkit/application/application.py
+++ b/prompt_toolkit/application/application.py
@@ -366,10 +366,10 @@ class Application(Generic[_AppResult]):
         depth = self._color_depth
 
         if callable(depth):
-            return depth() or self.output.get_default_color_depth()
+            depth = depth()
 
         if depth is None:
-            return self.output.get_default_color_depth()
+            depth = self.output.get_color_depth()
 
         return depth
 
diff --git a/prompt_toolkit/output/base.py b/prompt_toolkit/output/base.py
index 7d056678..8b91083e 100644
--- a/prompt_toolkit/output/base.py
+++ b/prompt_toolkit/output/base.py
@@ -164,8 +164,8 @@ class Output(metaclass=ABCMeta):
         raise NotImplementedError
 
     @abstractmethod
-    def get_default_color_depth(self) -> ColorDepth:
-        " Get default color depth for this output. "
+    def get_color_depth(self) -> ColorDepth:
+        " Get color depth for this output. "
 
 
 class DummyOutput(Output):
@@ -270,5 +270,5 @@ class DummyOutput(Output):
     def get_rows_below_cursor_position(self) -> int:
         return 40
 
-    def get_default_color_depth(self) -> ColorDepth:
+    def get_color_depth(self) -> ColorDepth:
         return ColorDepth.DEPTH_1_BIT
diff --git a/prompt_toolkit/output/defaults.py b/prompt_toolkit/output/defaults.py
index 2af7615d..6f4890da 100644
--- a/prompt_toolkit/output/defaults.py
+++ b/prompt_toolkit/output/defaults.py
@@ -4,6 +4,7 @@ from typing import Optional, TextIO, cast
 from prompt_toolkit.patch_stdout import StdoutProxy
 from prompt_toolkit.utils import (
     get_term_environment_variable,
+    get_color_depth_environment_variable,
     is_conemu_ansi,
     is_windows,
 )
@@ -45,18 +46,23 @@ def create_output(
     while isinstance(stdout, StdoutProxy):
         stdout = stdout.original_stdout
 
+    # Get color depth from the PROMPT_TOOLKIT_COLOR_DEPTH environment variable,
+    # None if the variable is not set.
+    color_depth = get_color_depth_environment_variable()
+
     if is_windows():
         from .conemu import ConEmuOutput
         from .win32 import Win32Output
         from .windows10 import is_win_vt100_enabled, Windows10_Output
 
         if is_win_vt100_enabled():
-            return cast(Output, Windows10_Output(stdout))
+            return cast(Output, Windows10_Output(stdout, color_depth=color_depth))
         if is_conemu_ansi():
-            return cast(Output, ConEmuOutput(stdout))
+            return cast(Output, ConEmuOutput(stdout, color_depth=color_depth))
         else:
-            return Win32Output(stdout)
+            return Win32Output(stdout, color_depth=color_depth)
     else:
         from .vt100 import Vt100_Output
 
-        return Vt100_Output.from_pty(stdout, term=get_term_environment_variable())
+        term = get_term_environment_variable()
+        return Vt100_Output.from_pty(stdout, term=term, color_depth=color_depth)
diff --git a/prompt_toolkit/output/vt100.py b/prompt_toolkit/output/vt100.py
index b71a3c2f..be7a5c1e 100644
--- a/prompt_toolkit/output/vt100.py
+++ b/prompt_toolkit/output/vt100.py
@@ -415,6 +415,7 @@ class Vt100_Output(Output):
         get_size: Callable[[], Size],
         term: Optional[str] = None,
         write_binary: bool = True,
+        color_depth: Optional[ColorDepth] = None,
     ) -> None:
 
         assert all(hasattr(stdout, a) for a in ("write", "flush"))
@@ -427,6 +428,7 @@ class Vt100_Output(Output):
         self.write_binary = write_binary
         self._get_size = get_size
         self.term = term
+        self._color_depth = color_depth
 
         # Cache for escape codes.
         self._escape_code_caches: Dict[ColorDepth, _EscapeCodeCache] = {
@@ -437,7 +439,12 @@ class Vt100_Output(Output):
         }
 
     @classmethod
-    def from_pty(cls, stdout: TextIO, term: Optional[str] = None) -> "Vt100_Output":
+    def from_pty(
+        cls,
+        stdout: TextIO,
+        term: Optional[str] = None,
+        color_depth: Optional[ColorDepth] = None,
+    ) -> "Vt100_Output":
         """
         Create an Output class from a pseudo terminal.
         (This will take the dimensions by reading the pseudo
@@ -470,7 +477,7 @@ class Vt100_Output(Output):
                 pass
             return Size(rows=rows or 24, columns=columns or 80)
 
-        return cls(stdout, get_size, term=term)
+        return cls(stdout, get_size, term=term, color_depth=color_depth)
 
     def get_size(self) -> Size:
         return self._get_size()
@@ -684,5 +691,7 @@ class Vt100_Output(Output):
         self.write_raw("\a")
         self.flush()
 
-    def get_default_color_depth(self) -> ColorDepth:
+    def get_color_depth(self) -> ColorDepth:
+        if self._depth is not None:
+            return self._depth
         return ColorDepth.vt100_default(self.term)
```

This would mean the color depth is defined when creating the output object, but can be overridden by the application.

To sum it up:
- for `create_output()`, `get_color_depth()` defaults to:
     * `PROMPT_TOOLKIT_COLOR_DEPTH` if provided
     * else computed from output type and possibly `output.term`
- for `Vt100_Output`, `Win32Output`, etc. it is:
     * provided through the `color_depth` argument if provided
     * else computed from output type and possibly `output.term`
- for `Application`, the color depth can be:
     * overridden using the `color_depth` argument
     * defaulting to `self.output.get_color_depth()` otherwise 

Would that make sense? It seems less painful to me than adding the `ignore_environment_variables` argument (especially since it's only used for `PROMPT_TOOLKIT_COLOR_DEPTH`), but I might be missing something :)

jonathanslenders:
Thanks! Very good feedback.

Quick question: Some applications like ptpython allow the user to configure the color depth in the application itself, making it possible to toggle between color depths at runtime. If a PROMPT_TOOLKIT_COLOR_DEPTH environment variable is set, should it get priority over the application setting or not? Right now, it's not, it is the default that's used if the application doesn't override it.

jonathanslenders:
See: https://github.com/prompt-toolkit/python-prompt-toolkit/pull/1184
 -> this new PR incorporates your suggestion. I did not yet test it.

vxgmichel:
> If a PROMPT_TOOLKIT_COLOR_DEPTH environment variable is set, should it get priority over the application setting or not? Right now, it's not, it is the default that's used if the application doesn't override it.

Following my comment in PR #1184: https://github.com/prompt-toolkit/python-prompt-toolkit/pull/1184#issuecomment-653076526

If `PROMPT_TOOLKIT_COLOR_DEPTH` is defined as:
> set the supported color depth for the local terminal

Then I'd say the application color depth has priority over it as it only indicated the "supported" color depth. As I mentioned in my comment, a warning can be issue if the application is configured with a color depth higher than the supported color depth for the provided output.

PR #1184, opened 2020-07-01
Move colordepth logic to Output implementations and added ignore_envi…

jonathanslenders:
…ronment_variables parameter to AppSession.

### DIFF
--- a/prompt_toolkit/application/application.py
+++ b/prompt_toolkit/application/application.py
@@ -366,15 +366,23 @@ class Application(Generic[_AppResult]):
     @property
     def color_depth(self) -> ColorDepth:
         """
-        Active :class:`.ColorDepth`.
+        The active :class:`.ColorDepth`.
+
+        The current value is determined as follows:
+        - If a color depth was given explicitely to this application, use that
+          value.
+        - Otherwise, fall back to the color depth that is reported by the
+          :class:`.Output` implementation. If the :class:`.Output` class was
+          created using `output.defaults.create_output`, then this value is
+          coming from the $PROMPT_TOOLKIT_COLOR_DEPTH environment variable.
         """
         depth = self._color_depth
 
         if callable(depth):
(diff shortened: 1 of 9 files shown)


vxgmichel:
If I understand it correctly, the logic is:

- `application.color_depth`
   * provided as argument, defaulting to `output.get_default_color_depth()`
- `output.get_default_color_depth()`:
   * provided as argument, defaulting to an output-specific value (that might depend on `output.term`)

It looks pretty good, I was just a bit confused by the fact that a `default_color_depth` can be customized. It probably comes from that fact that we have 3 levels of color depth: `color_depth`, `default_color_depth` and `default_default_color_depth`.

I suppose the alternative would be a slightly different naming:
- `application.color_depth`
- `output.get_supported_color_depth()`
- `output.get_default_supported_color_depth()`

This might make a bit more sense if we end implementing this kind of logic:
```python
    # Application.color_depth
    @property
    def color_depth(self):
        supported_color_depth = self.output.get_supported_color_depth()
        if self._color_depth is None:
             return supported_color_depth
        if self._color_depth > supported_color_depth:
            warning.warns("Color depth is not supported by ouput")
            return supported_color_depth
        return self._color_depth
```

I don't have a strong opinion about it though :)

vxgmichel:
LGTM otherwise! 

Another comment is that I would probably move `ColorDepth.from_env` to `prompt_toolkit.utils.get_color_depth_from_env` so all the environment variable helpers are gathered in the same module.

This way it'd be easier to track the use of environment variable to make sure they don't affect remote terminals.

jonathanslenders:
I did some little changes, and will merge it like this for now.

``ColorDepth.from_env`` had to stay because of backwards-compatibility. (ptpython would break for instance).

This check here what you propose:
```python
        if self._color_depth > supported_color_depth:
            warning.warns("Color depth is not supported by output")
```
that makes sense. I guess I can do this in a separate PR.

Issue #192, opened 2015-11-18
cursor shape (e.g. for vim mode)


davidszotten:
hey, thanks for an awesome library!

would it be possible to support cursor shapes, like https://hamberg.no/erlend/posts/2014-03-09-change-vim-cursor-in-iterm.html (or http://vim.wikia.com/wiki/Change_cursor_shape_in_different_modes)

jonathanslenders:
Hey @davidszotten,

That is probably possible, yes. Although I need to have a look how to implement this cleanly [1] and making sure that most terminals understand it. It's not a priority for me right now, but I'll put it on the to-do list.

Jonathan

[1] The tricky part is that the Vi state (insert/navigation) is currently not available from the CLI/renderer and the state is not able to access the renderer. So, I'm not sure yet how to make it fit cleanly in the architecture.

sassanh:
@jonathanslenders What about dirty temporary workarounds for endusers? It's not healthy for vim users to work with wrong cursor shape 😉
Is there anything like a signal fired when vim mode is changed so that we can print the appropriate character sequence to terminal that it changes the cursor shape? I know the character sequences that change cursor shape for my terminal, so if we have those signals and a way to print to terminal when those signals are fired we can have temporary workarounds for our setups.

jonathanslenders:
Hi @sassanh, I understand that this is useful. But so far I did not had time yet (actually it wasn't a priority). If you have a patch, then I'm more than happy to review it. Otherwise, I'll have a look at this later on myself. Probably it needs to go somewhere in the Renderer class, but I'm not sure what's the best approach.

jonathanslenders:
Other question: is there somewhere some official documentation about escape sequences for cursor shapes? Which terminals support it? Do all terminals use the same escape sequences? And if not, how do I know which escape sequences to use?

sassanh:
@jonathanslenders I understand, no rush. Just wanted to know if I can use a temporary hacky solution to acheive this till you have time to implement it.
[...]

PR #1195, opened 2020-07-21
Introduce vi_state.InputMode.REPLACE_SINGLE

bitkeen:
This would allow to customize cursor shape for single-character replace
mode, similar to how t_SR works in Vim (both for r and R).

See https://github.com/prompt-toolkit/python-prompt-toolkit/issues/192 for examples of cursor-shape customization.


jonathanslenders:
Thanks a lot! Perfect!

Issue #1157, opened 2020-06-05
Shift-Selection not working; bug in unshift_move


scoennz:
Hi,
I have just installed Winpython 3.8.3.0, which ships with IPython 7.14 based on prompt_toolkit 3. I have just noticed that the shift-selection mode does not work in the IPython terminal. Nothing happens when doing shift-left or shift-right, eg (this should start a selection). I have traced that issue to the following code in the `unshift_move(..)` function in `emacs.py` - this function transforms eg a Shift-Left event into a Left event and call the handler, to move the cursor after starting the selection:

https://github.com/prompt-toolkit/python-prompt-toolkit/blob/5f2268d9232f5d300b4a3226c6e82fe44c60f9ca/prompt_toolkit/key_binding/bindings/emacs.py#L439-L448

It turns out the 'handler' is actually an instance of Binding. Inserting a `print(handler)` statement in the code and typing Shift-Left eg displays:

`Binding(keys=(), handler=<function backward_char at 0x000001A3F5B409D0>)`

As a result, handler is never called because contrary to what is commented `handler` is not just a normal callable (I can't figure why not, though; this seems to have changed since I last looked at this). Removing `if not isinstance(handler, Binding):` and calling handler.handler(event) solves the issue for me. But perhaps there is a cleaner way of fixing this? 

Thanks for looking into this. Best regards, Stephane

(I noticed that your profile says that you are based in Belgium. I am from Belgium too! But have been living in NZ for quite a while now.)

WhistleWhileYouWork:
I can confirm that as of `prompt_toolkit` 3.0.4 -> 3.05, CTRL+SHIFT and SHIFT selection are broken in Windows Console using any combination of iPython 7.10 -> 7.15 with Python 3.7 -> 3.8.3. Rolling back to `prompt_toolkit` 3.0.3 resolves the issue.

Thank you.

scoennz:
Found the issue. `named_commands.py` was modified in commit 148833e982cc0ed427a862e0d2725b353af39d04 on Feb 3 so that the registering of commands only procuce `Bindings`. See excerpt from code below:

https://github.com/prompt-toolkit/python-prompt-toolkit/blob/5f2268d9232f5d300b4a3226c6e82fe44c60f9ca/prompt_toolkit/key_binding/bindings/named_commands.py#L41-L45

To take that into account in the `unshift_move(..)` function in `emacs.py`, the try-except block needs just be modified like this:

```python
[...]

PR #1212, opened 2020-08-16
Tiny fix to emacs shift selection

dannysepler:
Considering trying to import emacs shift selection into xonsh. It looks like this handler is unreachable at the moment, since `get_by_name` always returns a binding.


prompt_toolkit/key_binding/bindings/emacs.py
```diff
@@ -438,10 +438,10 @@ def unshift_move(event: E) -> None:
 
         try:
             # Both the dict lookup and `get_by_name` can raise KeyError.
-            handler = get_by_name(key_to_command[key])
+            handler = get_by_name(key_to_command[key]).handler
         except KeyError:
             pass
-        else:  # (`else` is not really needed here.)
+        else:  # (We don't want the handler to execute if there was a KeyError)
```
dannysepler, line 444: This corresponds to the comment given in this diff https://github.com/prompt-toolkit/python-prompt-toolkit/pull/989/files#r344139334

dannysepler:
this should fix https://github.com/prompt-toolkit/python-prompt-toolkit/issues/1157 -- i also like how the code was written there better than my original idea

PR #1215, opened 2020-08-17
Added 'placeholder' parameter to PromptSession.

jonathanslenders:


Yajo:
When placeholder contains `\n`, it outputs `^J`... is that normal?

PR #1231, opened 2020-09-02
Alias escape with shift in some emacs key bindings

dannysepler:
I use a mac, and my fingers like Option more than Control (since it's slightly closer to my thumb). Most text editors alias Ctrl-Left with Option-Left, might be nice if PTK did the same.

Upon rereading, I'm a tad worried I'm missing nuances between Ctrl and Option, lemme know if I'm just imagining them!

### DIFF
--- a/prompt_toolkit/input/ansi_escape_sequences.py
+++ b/prompt_toolkit/input/ansi_escape_sequences.py
@@ -267,6 +267,8 @@ ANSI_SEQUENCES: Dict[str, Union[Keys, Tuple[Keys, ...]]] = {
     "\x1b[1;9B": (Keys.Escape, Keys.Down),
     "\x1b[1;9C": (Keys.Escape, Keys.Right),
     "\x1b[1;9D": (Keys.Escape, Keys.Left),
+    "\x1b[1;10C": Keys.EscapeShiftRight,
+    "\x1b[1;10D": Keys.EscapeShiftLeft,
     # --
     # Control/shift/meta + number in mintty.
     # (c-2 will actually send c-@ and c-6 will send c-^.)
--- a/prompt_toolkit/input/ansi_escape_sequences.py
+++ b/prompt_toolkit/input/ansi_escape_sequences.py
@@ -267,8 +267,8 @@ ANSI_SEQUENCES: Dict[str, Union[Keys, Tuple[Keys, ...]]] = {
     "\x1b[1;9B": (Keys.Escape, Keys.Down),
     "\x1b[1;9C": (Keys.Escape, Keys.Right),
     "\x1b[1;9D": (Keys.Escape, Keys.Left),
-    "\x1b[1;10C": Keys.EscapeShiftRight,
-    "\x1b[1;10D": Keys.EscapeShiftLeft,
+    "\x1b[1;10C": (Keys.Escape, Keys.ShiftRight),
+    "\x1b[1;10D": (Keys.Escape, Keys.ShiftLeft),
     # --
     # Control/shift/meta + number in mintty.
     # (c-2 will actually send c-@ and c-6 will send c-^.)
--- a/prompt_toolkit/key_binding/bindings/emacs.py
+++ b/prompt_toolkit/key_binding/bindings/emacs.py
@@ -1,5 +1,5 @@
 # pylint: disable=function-redefined
-from typing import Dict, Union
+from typing import Dict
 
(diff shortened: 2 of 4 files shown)


prompt_toolkit/key_binding/bindings/emacs.py
```diff
@@ -269,27 +271,6 @@ def _copy(event: E) -> None:
         data = event.current_buffer.copy_selection()
         event.app.clipboard.set_data(data)
 
-    @handle("escape", "left")
```
dannysepler, line 272: could be mistaken, but these functions seem the same as `backward-word` and `forward-word`?

prompt_toolkit/input/ansi_escape_sequences.py
```diff
@@ -267,6 +267,8 @@
     "\x1b[1;9B": (Keys.Escape, Keys.Down),
     "\x1b[1;9C": (Keys.Escape, Keys.Right),
     "\x1b[1;9D": (Keys.Escape, Keys.Left),
+    "\x1b[1;10C": Keys.EscapeShiftRight,
```
jonathanslenders, line 270: Hi @dannysepler,

These changes look fine for me, except for one thing. "escape" is always considered a separate key in prompt_toolkit, unlike control or shift. The reason is that people can type "escape" first, and later the arrow key, or they can do "alt" + arrow key. For the parser, the escape is a separate key.

So, this means that like the above key bindings, this should be a tuple (an ANSI sequence that corresponds with multiple keys):
```python
    "\x1b[1;10C": (Keys.Escape, Keys.ShiftRight)
```

prompt_toolkit/key_binding/bindings/emacs.py
jonathanslenders: It looks like you're right for ``backward-word``.

For ``forward-word`` the implementation is not the same. Are you sure they are doing the same thing? I mean, one does ``find_next_word_ending`` and the other does ``find_next_word_beginning``.

I'm not sure what is right. I'm all for less code duplication and doing the same thing GNU readline or Emacs is doing.

prompt_toolkit/key_binding/bindings/emacs.py
```diff
@@ -433,6 +414,8 @@ def unshift_move(event: E) -> None:
             Keys.ControlShiftRight: "forward-word",
             Keys.ControlShiftHome: "beginning-of-buffer",
             Keys.ControlShiftEnd: "end-of-buffer",
+            Keys.EscapeShiftLeft: "backward-word",
```
jonathanslenders, line 417: So, here we should turn it into a ``Dict[Tuple[Union[Keys, str], ...], str]`` (keys become tuples). and extract that from the key_sequence.

prompt_toolkit/keys.py
```diff
@@ -120,6 +120,9 @@ class Keys(str, Enum):
     ControlShiftPageUp = "c-s-pageup"
     ControlShiftPageDown = "c-s-pagedown"
 
+    EscapeShiftLeft = "e-s-left"
```
jonathanslenders, line 123: These we don't need.

jonathanslenders:
Thanks a lot @dannysepler, this looks like a good contribution. See the remarks for a few things that require changes.

prompt_toolkit/key_binding/bindings/emacs.py
dannysepler: great point! the [docs](https://www.gnu.org/software/emacs/manual/html_node/emacs/Words.html) were a tad hard to understand this particular part. so i downloaded emacs to see for myself. What's @register-ed in "named_commands" was right. specifically:

- `backward-word` goes to the start of the last word
- `forward-word` goes to the next non-letter-or-number (space, punctuation, etc)

prompt_toolkit/key_binding/bindings/emacs.py
```diff
@@ -414,33 +395,35 @@ def unshift_move(event: E) -> None:
         a shift + movement key press event, moves the cursor
         as if shift is not pressed.
         """
-        key = event.key_sequence[0].key
+        sequence = " ".join([seq.key for seq in event.key_sequence])
```
dannysepler, line 398: using tuples made all this code surprisingly painful, so i took some creative liberty. but do lmk if you'd rather the tuples, and i'll bring em back!

Issue #1211, opened 2020-08-12
"radiolist focused" versus "radio-list focused" or perhaps a bug in RadioList focusing?


JoshHeitzman:
I'm not sure if this is a question or a bug report.  The file examples/full-screen/full-screen-demo.py has an app style that includes:
        "radiolist focused": "noreverse",
        "radiolist focused radio.selected": "reverse",

Where as the prompt_toolkit.widgets.base.RadioList class has:
    container_style = "class:radio-list"
    default_style = "class:radio"
    selected_style = "class:radio-selected"
    checked_style = "class:radio-checked"

That container_style's use of 'radio-list' rather than 'radiolist' has me wondering if the example's code should include a hyphen.  I tried changing example to use 'radio-list' instead of 'radiolist on those two lines, but I didn't see any visible difference.  I did some debugging in prompt_toolkit.styles.style.Style.get_attrs_for_style_str and didn't observe a match occur that resulted in that function's local variable list_of_attrs being appended too.  I tried that debuggin with both "radiolist focused" and "radio-list focused".  But, since the only combos I observed that included 'radio' did not included 'focused', perhaps focused isn't handled here.  This is my first delve into styles, so I'm not sure were all of the relevant code is, and very well may have missed something.

In turn, I'm not certain if I'm asking "How is this supposed to work?" or reporting that something isn't working or that there is some superfluous code in that sample.

jonathanslenders:
This sounds like a bug.

Two bugs actually:
- radio-list needs a dash everywhere.
- "focused" is not anymore attached as a style to the focused region. I guess (but don't remember 100%), because it was not working in some situations (like where the focus is supposed to be not a single ``Window``, but a parent ``Container``, or when there are floating containers on top). Some widgets, like ``Button`` however, do expose a style like ``button.focused``.\

For more information, also see: https://python-prompt-toolkit.readthedocs.io/en/master/pages/advanced_topics/styling.html

PR #1244, opened 2020-09-14
Remove old invalid style rules in full-screen-demo.py example.

jonathanslenders:


jonathanslenders:
See: https://github.com/prompt-toolkit/python-prompt-toolkit/issues/1211

PR #1255, opened 2020-09-26
Fix race condition in `ProgressBar` shutdown logic

### LINKED ISSUE #964: Race condition in ProgressBar
There appears to be a race condition in the `ProgressBar`'s context manager logic where if an exception is raised (or, indeed, if execution finishes) before the `app` has started, `self.app.is_running` will be false, the `app` will not be exited, and the thread will be left to run indefinitely.  You can reproduce this rather trivially (assuming starting a new thread takes as long as it does for me!) with:

```py

with ProgressBar():
    pass
```

layday:
The app might not have started upon `__exit__`ing
the context manager from the main thread;
in which case `self.app.exit()` has no effect,
the app will be left to run indefinitely
and the program will hang after `join`ing
the app thread.
Instead, we can repurpose `on_invalidate`
to call `app.exit` the next time
the app is invalidated.
This relies on the app's `refresh_interval`
being set to a value other than `None` for it
to happen automatically.

Closes #964.

### DIFF
--- a/prompt_toolkit/shortcuts/progress_bar/base.py
+++ b/prompt_toolkit/shortcuts/progress_bar/base.py
@@ -50,7 +50,7 @@ from prompt_toolkit.layout.controls import UIContent, UIControl
 from prompt_toolkit.layout.dimension import AnyDimension, D
 from prompt_toolkit.output import ColorDepth, Output
 from prompt_toolkit.styles import BaseStyle
-from prompt_toolkit.utils import in_main_thread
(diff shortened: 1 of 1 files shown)


layday:
I can't really think of a way to test this unless we are to factor out `run` and pass it to `ProgressBar` as an argument.  There might be a better way to signal to the app that it should exit, but `async_run` and the accompanying logic wasn't easy to follow.

PR #1259, opened 2020-09-29
Deterministic minimun size for completion menu, even at bottom of screen (for `PromptSession`)

bobhy:


bobhy:
When using `PromptSession` shortcut (and friends) and displaying the completion menu, if there are "many" rows of completions, the number of rows actually displayed depends on height of the prompt line(s) above the bottom of the screen: 
near the top of the screen, the whole completion menu is displayed, but at the bottom, a hardcoded minimum of 3 lines is displayed  Option `reserve_space_for_menu` was intended to allow an application to ensure more than 3 lines if desired, but it is not working.

In addition, for a multi-column completion menu with 3 available completions (same as hard-coded height above), the completion menu would be displayed as 3 rows of 1 column, rather than as 1 row of 3 columns.  This causes unnecessary scrolling when the prompt is at the bottom of the screen (which it usually is).

And finally, for a `MultiColumnCompletionMenu` with `show_meta == True` (the default), the meta row was always being displayed, even if no completion had any meta display data, again squandering that precious bottom-of-the-screen real estate.  

This PR does the following:
1. Reimplements `reserve_space_for_menu`, so (at least) the specified number of rows of completions is displayed anyhere on the screen.  It uses a new option for `FloatContainer` to include preferred heights of the child `Float` containers as well as the background container itself in the calculation of its preferred height for the `FloatContainer`. ( see `floats_min_preferred_height`).  The preferred height will be the actual number of completion rows, or the configured limit, whichever is smaller.
2. In `class Completion`, avoid defaulting meta to `""` actual value is None.  This was breaking code already in `MultiColumnCompletionMenu' that was trying to hide meta display line if no completion actually had meta. 
3. in `PromptSession` layout, fixed right prompt so it does not add to the height of the prompt if it is not configured.  Once the floats for completion menu and right prompt are included in preferred height calculation, it is important that they report preferred height of 0 when not in use.  This doesn't avoid an unnecessary blank line when right prompt *is* configured, however. 
4. Lay out the right prompt to start on the same row as the main (left) prompt.  It was floating all the way to the bottm of the screen even if the prompt was at the top!

bobhy:
Superceded by #1271.

PR #1262, opened 2020-10-02
Fix: Typo and whitespace

farisachugthai:
Needed whitespace for the docstring to render correctly after running
the project through sphinx. As it stands, the current `-` renders on the
same line as the previous one instead of starting a unordered list.

Also fix a misspelling.


jonathanslenders:
Thanks!

PR #1271, opened 2020-10-11
Completion menu rows - Clean up multi-column completion display, especially predictable height

bobhy:
Replace `PromptSession` parameter `reserve_space_for_menu`, which wasn't really working, with a new parameter `completion_menu_rows` that has slightly different semantics.  Additional layout fixes affecting both `CompletionMenu` and `MultiColumnCompletionMenu` noted below.

Parameter `completion_menu_rows` limits the number of rows that a completion menu will display, default 5.  It applies equally to `CompletionMenu` and `MultiColumnCompletionMenu`.  The point of putting a limit on number of rows is that it will render faster and users  wouldn't read through a really long completion list anyway, they'll type a few more characters to narrow it down.  Other changes ensure that the number of rows will not be reduced when at/near the bottom of the screen.  The parameter can be specified on `PromptSession.prompt{()` or `PromptSession.prompt_async()` as well.  Old parameter `reserve_space_for_menu` produces a `DeprecationWarning` if specified, though the value is used for the new behavior.

Other changes in the PR:
* `completion.Completion.display_meta_text` returns `None`, not empty string if Completion has no meta data.  This prevents completion menus from displaying an extra blank line if no completion had meta data.
* `containers.Float` -- new option `dont_shrink_height: bool = False`. If `True`, height of Float will not be reduced if it wouldn't fit in `FloatContainer`.  Instead, its preferred height will be considered in the preferred height of the `FloatContainer`. Default is `False`, the prior behavior.  PromptSession layout uses this to ensure completion menus are not shrunk at bottom of screen.
* `layout.MultiColumnCompletionMenu` -- new option `max_rows: int = 5` specifies an upper limit on number of rows in the completion menu. Similar to effect of `max_height` in `layout.CompletionMenu`.  Also changed the column width calculation to handle outliers in length of completions more gracefully.


bobhy:
Rebased on latest master.  Added missing backward compatibility: `reserve_space_for_menu` can now be specified (but will provide the deprecation warning, as advertised).

ju-kreber:
This works really well for me, an issue with the completion dialogue is gone and the behavior of space reservation is improved.

t184256:
Can confirm that this PR solves https://github.com/xonsh/xonsh/issues/3810 and makes `multi` tab completion ellipsize long entries.

Spriithy:
Any chance this MR might be merged for the next release ?

anki-code:
Hi @jonathanslenders! What do you think about this PR? Looks like many thumb ups here.

PR #1278, opened 2020-10-24
Docs rewording

byrro:
Clarification on argument and method available.

Couple suggestions to rewrite the docs in a more clear way.

Good portion is quite subjective. Probably too obvious, but anyway, feel free to judge the merit of the suggestions.


jonathanslenders:
Thanks!

PR #1279, opened 2020-10-24
App exitable argument for Application object

byrro:
Making an app exitable is commonly needed (almost all fullscreen apps will want it?). Being able to set this key-binding right when an `Application` object is being created is handy:

Quick examples: `app = Application(exit_on_key="c-q")` or `app = Application(exit_on_key=("escape", "q"))`.

This would also simply some examples in the docs that are currently unexit'able (quite annoying to run, actually).

Let me know if this makes sense or not. Open to suggestions for changing anything also.

This feature does not change the default behavior of the `Application` object (no key-bindings). It's also overriden by the currently available `key_bindings` argument.

Wrote four tests which I think covers most common scenarios. Didn't see the need to write more, but let me know if I missed anything important.

### DIFF
--- a/prompt_toolkit/application/__init__.py
+++ b/prompt_toolkit/application/__init__.py
@@ -1,4 +1,4 @@
-from .application import Application
+from .application import Application, handle_exit_on_key
 from .current import (
     AppSession,
     create_app_session,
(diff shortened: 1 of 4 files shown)


byrro:
I'm running `isort` locally on py36-38 against `tests/test_application.py`, it doesn't point any errors, but on TravisCI it keeps complaining. It shows a deprecation and update warning related to isort, perhaps it's an old version in the build server? At this point, I'm unsure what to do to get it passed...

PR #1291, opened 2020-11-22
Prevent asyncio DeprecationWarning

carmenbianca:
Under some circumstances (see <https://github.com/xonsh/xonsh/issues/3907>) a DeprecationWarning would be printed. This is because of passing coroutine objects directly to asyncio.wait(), which is confusing behaviour that has been deprecated.

<https://github.com/python/cpython/pull/9543> gives information on how to refactor code that uses the deprecated method.

Fixes https://github.com/xonsh/xonsh/issues/3907

### DIFF
--- a/prompt_toolkit/renderer.py
+++ b/prompt_toolkit/renderer.py
@@ -2,7 +2,7 @@
 Renders the command line on the console.
 (Redraws parts of the input line that were changed.)
 """
-from asyncio import FIRST_COMPLETED, Future, sleep, wait
+from asyncio import FIRST_COMPLETED, Future, create_task, sleep, wait
 from collections import deque
 from enum import Enum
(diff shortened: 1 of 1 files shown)


carmenbianca:
Ah… Broken under 3.6. [asyncio.create_task](https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task) was only introduced in 3.7.

Not sure what the preferred method of fixing this is. This could possibly do it if I'm not mistaken:

```python
try:
    from asyncio import create_task
except ImportError:
    create_task = lambda x: x
```

But that's obviously a bit hacky. Happy to take input :)

jonathanslenders:
Can we use ``asyncio.ensure_future()`` instead of ``asyncio.create_task()``?

carmenbianca:
@jonathanslenders That appears to work, although the documentation for ensure_future() is esoteric magic to me.

jonathanslenders:
Thank you @carmenbianca !

Issue #1294, opened 2020-11-27
s-tab support for TERM linux


drewarnett:
Wrote an application I want to be as portable as possible.  Found that s-tab didn't work in the linux console (TERM=linux).  Found a great article talking about how linux doesn't match ANSI/VT for everything.  (TODO:  find that and add a link here.)

I'll provide a pull request that will simply add an additional escape sequence to prompt_toolkit.input.ansi_escape_sequences that works.  (Rename to .py.  Import.  Call patch() within.)
[patch_for_linux.txt](https://github.com/prompt-toolkit/python-prompt-toolkit/files/5605470/patch_for_linux.txt)

Until a fix is incorporated and released, I'm using the attached file as a workaround, as I'd like to not have an application dependency on a non-released (ie, pypi) version of prompt-toolkit.

Putting up this issue and not just submitting a pull request for a couple of reasons.

First, is the patch in my pull request the best way to implement this?  Shouldn't break anything as far as I know, and seems to already be TERM=linux alternative sequences in there already.

Second, any value to having prompt-toolkit read TERM and build ANSI_SEQUENCES dynamically based on that?

drewarnett:
Mistake in first pull request, so closed that and did a new, correct pull request.

PR #1295, opened 2020-11-27
add BackTab sequence for TERM=linux

drewarnett:
Offered as solution for #1294.


drewarnett:
Forgot comma.  Will fix and submit new pull request.

Issue #363, opened 2016-07-26
Option to disable bell?


chris-b1:
Thanks for the great library.  I'm using ConEmu on Windows, and one thing that's a little annoying upgrading to IPython 5.0 is the bell sound that is triggered if I hit backspace on a an empty line - just quickly skimming the code - I'm assuming it's triggered here? 

https://github.com/jonathanslenders/python-prompt-toolkit/blob/master/prompt_toolkit/key_binding/bindings/named_commands.py#L181

I don't really know how the configuration of prompt toolkit works, but ideally there would be some way to turn this off that could be exposed through IPython?

chris-b1:
So I upgraded ConEmu, and there actually is now an option to disable the bell there.  I'll leave this open, but not a priority, or feel free to close.

https://conemu.github.io/en/SettingsFeatures.html

asmeurer:
I don't know about IPython, but for a general pt shell but you could disable the bell entirely by subclassing ConEmuOutput, stubbing out `bell`, and using that as the `output` argument for `CommandLineInterface`. IPython makes it impossible or next to impossible to make these kinds of changes to the pt classes in my experience, so it likely would be difficult to get it working there.

rsrini7:
Depends on Emulators settings. Same issue with ptpython.
To disable bell sound in Windows Terminal -> defaults -> add  "bellStyle":"none"

PR #1333, opened 2021-01-21
Disable bell with environment variable

daskol:
I have that annoying beep on editing read-only buffer. Blacklisting `pcspkr` kernel module is not an option for me (it has global effect). This PR allows to configure beeping with `PROMPT_TOOLKIT_BELL` environment variable. Probably event better solution would be parsing `.inputrc` for bell configuration. Run the following for testing.

```shell
PROMPT_TOOLKIT_BELL=false ipython  # turn off beeping
```

Related issues are #363, ipython/ipython#11626, and ipython/ipython#12196.

### DIFF
--- a/prompt_toolkit/output/vt100.py
+++ b/prompt_toolkit/output/vt100.py
@@ -488,7 +488,13 @@ class Vt100_Output(Output):
                 pass
             return Size(rows=rows or 24, columns=columns or 80)
 
-        return cls(stdout, get_size, term=term, default_color_depth=default_color_depth, bell=bell)
+        return cls(
+            stdout,
+            get_size,
+            term=term,
+            default_color_depth=default_color_depth,
+            bell=bell,
+        )
 
     def get_size(self) -> Size:
         return self._get_size()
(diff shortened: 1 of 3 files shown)


prompt_toolkit/output/vt100.py
```diff
@@ -432,6 +433,7 @@ def __init__(
         self.default_color_depth = default_color_depth
         self._get_size = get_size
         self.term = term
+        self.beep = bell
```
jonathanslenders, line 436: Can you call this ``enable_bell``? Both for the attribute and the parameter?
(I see it's called ``beep`` here to avoid collision with the method.)

jonathanslenders:
Thanks!

eugenesvk:
@daskol Am I correct in my understanding that this PR only adds `enable_bell=bell_from_env,` in `defaults.py` to non-Windows terminals? Is there a way to also extend this to Windows Terminal as the bell sound on backspace there is just as annoying and there isn't yet a terminal command to disable it :( 
Thanks!

daskol:
@eugenesvk Yes, that's right. There is nothing special to support the same feature for Windows. The only reason why it is not fixed in this pull request is that I have no Windows station for testing. Since Windows console emulator is based on Linux one it should be quite easy to manage ring for Windows terminal.

PR #1343, opened 2021-01-26
Buffer._set_cursor_position(): Fix return value for negative arguments

plotski:
I don't know if it's a bugfix, but this is more correct.


jonathanslenders:
Thanks!

PR #1355, opened 2021-02-07
Docs: fix misused automodule directive in docs/pages/reference.rst

MrMino:
TL;DR: "Reference" section of the docs: `.. autmodule::` → `.. autoclass::` where needed, add some headings to "Layout" to make it easier to navigate.

The [Layout](https://github.com/prompt-toolkit/python-prompt-toolkit/blob/91916569efb0e3c6709930c08760755d756844da/docs/pages/reference.rst#layout) section in the docs RST source uses `.. automodule::` multiple times, each time adding the same module docstring to the generated documentation. This makes for bugs in the TOC tree, menu JS, as well as anchors:

image

"Style" section had similar problem, but it was a lot less pronounced, since `prompt_toolkit.styles` module docstring is quite sparse at the moment.

This PR fixes it by replacing `.. autmodule::`  with  `.. autoclass::` where the former was duplicated, namely in "Style" and "Layout" sections. 

Headings under "Other" subsection of "Layouts" were added, to improve navigation.

See [Sphinx autodoc docs](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#directives) for reference.

### DIFF
--- a/docs/pages/reference.rst
+++ b/docs/pages/reference.rst
@@ -132,44 +132,116 @@ Lexers
 Layout
 ------
 
+.. automodule:: prompt_toolkit.layout
+
 The layout class itself
 ^^^^^^^^^^^^^^^^^^^^^^^
 
-.. automodule:: prompt_toolkit.layout
-    :members: Layout, InvalidLayoutError, walk
+.. autoclass:: prompt_toolkit.layout.Layout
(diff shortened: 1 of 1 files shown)


jonathanslenders:
Thanks!

PR #1356, opened 2021-02-07
Remove on_abort and on_exit from Application docstring

### LINKED ISSUE #1188: docs does not match implementation: Application(on_abort=..., on_exit=...)
the `on_abort` and `on_exit` parameters within `prompt_toolkit.application.Application` isn't available anymore in the latest code, but is still listed in the docs: https://github.com/prompt-toolkit/python-prompt-toolkit/blob/master/prompt_toolkit/application/application.py#L122

MrMino:
These parameters no longer exist in `Application.__init__` signature.

Fixes #1188.


jonathanslenders:
Thanks again for all the improvements!

PR #1358, opened 2021-02-07
Docs: remove mention of 'six' as required by PT

### LINKED ISSUE #1301: Documentation still lists six as required
The documentation still lists `six` as a requirement however it does not seem to be imported anymore.

PS: Additionally this library needs pygments but it is not listed in the requirements - is that an oversight or is there a reason for it?

MrMino:
Fixes #1301.

Prompt Toolkit does not require `six` anymore, but [the docs](https://python-prompt-toolkit.readthedocs.io/en/3.0.3/) still mention it as a dependency:

> Works everywhere: [...]
>  - Lightweight, the only dependencies are Pygments, **six** and wcwidth.


jonathanslenders:
Thanks!

PR #1366, opened 2021-02-16
Docs: make autodoc document Layout classes' methods

MrMino:
In #1355 I forgot to add `:members:` to each `.. autoclass::` directive, and so the methods went missing from the classes this was used on.

This PR fixes that.

For reference, see [sphinx.ext.autodoc docs](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#directive-autoclass):

> If you want to automatically document members, there’s a members option:
> (...)
> ```
> .. autoclass:: Noodle
>    :members:
> ```
> will document all non-private member functions and properties (that is, those whose name doesn’t start with _).
>


jonathanslenders:
Thanks!

PR #1378, opened 2021-02-21
Docs: Remove accidental getpass imports

JDeeth:
It looks like `import getpass` got copied a couple of times and included in examples where it isn't needed.


jonathanslenders:
Thanks!

PR #1382, opened 2021-02-25
add a filepath autocompleter to the examples

nathanrpage97:
Provide an autocomplete that is similar to IPython for files.

Will face issues with a few of the different edge cases (including spaces in files). But it gives the general idea of how to implement.

### DIFF
--- a/examples/prompts/auto-completion/autocomplete-file-path.py
+++ b/examples/prompts/auto-completion/autocomplete-file-path.py
@@ -10,7 +10,9 @@ _FILE_PATH_RE = re.compile(r"[^\s\"'\t]+")
 
 
 class PathCompleter(Completer):
-    def get_completions(self, document: Document, complete_event: CompleteEvent) -> Iterable[Completion]:
+    def get_completions(
+        self, document: Document, complete_event: CompleteEvent
+    ) -> Iterable[Completion]:
         path = document.get_word_before_cursor(pattern=_FILE_PATH_RE)
 
         # absoulte path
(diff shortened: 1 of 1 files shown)


jonathanslenders:
There is a ``PathCompleter`` built-in: https://github.com/prompt-toolkit/python-prompt-toolkit/blob/master/prompt_toolkit/completion/filesystem.py#L13
Is this different? Or can the example use that completer?

nathanrpage97:
oh wow, I couldn't find anything in the docs on this. Let me close this.

nathanrpage97:
Actually, the big difference is that this completer works at any "word" instead of the whole document text.

nathanrpage97:
Let me see if I can add a flag to the builtin PathCompleter that will enable it to be used from any position.

jonathanslenders:
If the path can appear at any "word" position, how is the output text handled by the application in that case?

In ptpython, ``PathCompleter`` is for instance combined with ``GrammarCompleter`` to allow path completion within quoted strings, (also automatically quoting/unquoting during completion). See: https://github.com/prompt-toolkit/ptpython/blob/master/ptpython/completer.py#L87 ``NestedCompleter`` also works nicely with ``PathCompleter`` I think. (https://github.com/prompt-toolkit/python-prompt-toolkit/blob/master/prompt_toolkit/completion/nested.py#L16)

I prefer different completers to be composable, rather then having all possible options and complexity in each completer. We could maybe split the ``WordCompleter`` in one completer that takes care of the tokenization and another one that holds the dictionary and is actually a ``SentenceCompleter`` (this latter does not exist, but that would be if ``sentence=True`` was passed to ``WordCompleter``). What you want is then to compose this new ``WordCompleter`` with the ``Pathcompleter``.

nathanrpage97:
Just saw this message. Would maybe providing the ptpython version, simplified to just the path component, be fine as an example of `AdvancedPathCompleter`?

Issue #1380, opened 2021-02-24
Allow binding Control+Backspace (at least in Windows)


eugenesvk:
One of my main usability issues with xonsh that's relying on your prompt toolkit is that I can't remap a very common key combo — <kbd>Control</kbd><kbd>Backspace</kbd> — to delete a previous word

There is already a similar [issue](https://github.com/prompt-toolkit/python-prompt-toolkit/issues/1250) that was closed with a reference to your [comment](https://github.com/prompt-toolkit/python-prompt-toolkit/issues/257#issuecomment-190328366); and as far as I understood, with the relevant code with some additional justification here:
https://github.com/prompt-toolkit/python-prompt-toolkit/blob/d6ac8be0791184d87262d5a759b11664dcdbf110/prompt_toolkit/input/ansi_escape_sequences.py#L58
>
    # Vt220 (and Linux terminal) send this when pressing backspace. We map this
    # to ControlH, because that will make it easier to create key bindings that
    # work everywhere, with the trade-off that it's no longer possible to
    # handle backspace and control-h individually for the few terminals that
    # support it. (Most terminals send ControlH when backspace is pressed.)
    # See: http://www.ibb.net/~anne/keyboard.html
    "\x7f": Keys.ControlH,

### Question 1
Is there any chance to revert this bundling of two keys into one key code?
The [blog](http://www.ibb.net/~anne/keyboard.html) mentioned above as part of the rationale for this seems to have changed its approach and (though I'm not certain) to now favor separating backspace and control-h so that "__^H ... can be used by applications__" (which I assume would allow mapping <kbd>Control</kbd><kbd>Backspace</kbd> separately from <kbd>Backspace</kbd>)

>
    In my previous attempt at working around the problem, I put most of my effort in trying to map all keys to suit xterm's vt100 emulation.
    However, I now propose another way to deal with the problem.
    This basically means getting xterm to work with the consoles' default settings: [<---] will do ASCII DEL (0x7F, or ^?), and [Delete] will do "\e[3~".
    This way ^H is not used for any terminal specific functions, it can be used by applications.

Or at least make this __optional/configurable__ so that a user can decide whether the trade-off is worth it?

### Question 2
Regardless of the general solution, is it possible to achieve the same in Windows, where `b"\x7f"` is already mapped to `Keys.Backspace`, not `Keys.ControlH`?
[...]

PR #1384, opened 2021-02-28
feat: separate handling of Control-H (^H) and Backspace

eugenesvk:
Allows achieving the console usability level of Windows :), where ^H and BS are already handled separately, e.g. with this patch you can bind `backward-kill-word` to Control-Backspace (which usually emits ^H) while leaving Backspace `backward-delete-char`

Per [this comment](https://github.com/prompt-toolkit/python-prompt-toolkit/issues/1380#issuecomment-786148973) this would be a breaking change, though a worthwile one

Closes the second part of #1380

### DIFF
--- a/prompt_toolkit/input/ansi_escape_sequences.py
+++ b/prompt_toolkit/input/ansi_escape_sequences.py
@@ -57,12 +57,11 @@ ANSI_SEQUENCES: Dict[str, Union[Keys, Tuple[Keys, ...]]] = {
     "\x1f": Keys.ControlUnderscore,  # Control-underscore (Also for Ctrl-hyphen.)
     # ASCII Delete (0x7f)
     # Vt220 (and Linux terminal) send this when pressing backspace. We map this
-    # to ControlH, because that will make it easier to create key bindings that
-    # work everywhere, with the trade-off that it's no longer possible to
-    # handle backspace and control-h individually for the few terminals that
-    # support it. (Most terminals send ControlH when backspace is pressed.)
+    # to Backspace, although that will make it harder to create key bindings that
+    # work everywhere given that some terminals don't support it
+    # (they send ControlH when backspace is pressed)
     # See: http://www.ibb.net/~anne/keyboard.html
-    "\x7f": Keys.ControlH,
+    "\x7f": Keys.Backspace,
     # --
     # Various
     "\x1b[1~": Keys.Home,  # tmux
--- a/prompt_toolkit/key_binding/bindings/basic.py
+++ b/prompt_toolkit/key_binding/bindings/basic.py
@@ -146,6 +146,9 @@ def load_basic_bindings() -> KeyBindings:
     handle("backspace", filter=insert_mode, save_before=if_no_repeat)(
         get_by_name("backward-delete-char")
     )
+    handle("c-h", filter=insert_mode, save_before=if_no_repeat)(
+        get_by_name("backward-delete-char")
+    )
     handle("delete", filter=insert_mode, save_before=if_no_repeat)(
         get_by_name("delete-char")
     )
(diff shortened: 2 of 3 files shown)


jonathanslenders:
Thanks. This sounds fine. If you have time, can you have a look at why the tests are failing? Otherwise I'll have a look later on.

eugenesvk:
Hey, I've already looked into it — please check this comment in the original issue https://github.com/prompt-toolkit/python-prompt-toolkit/issues/1380#issuecomment-787495825
Basically, we need to either remove the two tests or add the code to make the new keybind behave the same way as before the separation

> I see that there are two items in the test_cli.py failing, so if we don't want to add handle("c-h", filter=insert_mode)(get_by_name("backward-kill-word")) to emacs.py to make them pass I guess those need to be removed as Esc+^H wouldn't be expected to backward-kill-word like Esc+BS does?

Personally I'd prefer to add the keybinds so that `^H` kills a word everywhere (and that's what I'm doing in my xonsh configs), this would make the tests passing, but I

> ... left the non-Windows `^H` bound to `backward-delete-char`, not to `backward-kill-word` as I'm not sure if it'd break anything there

and need someone more knowledgeable about the use cases of ptk to make that call

jonathanslenders:
I'm trying the behavior in different terminals.
Putty seems to have a setting where backspace always sends ^H.

image

This means, that users with this setting enabled, won't be able to use backspace as usual if we bind ^H to backward-kill-word.

jonathanslenders:
What I'm also afraid of is that some users have the habit of literally typing control-h instead of backspace when they're in a terminal, because the key is easier to reach. So, I'm not sure we should make it a default on Linux. I'm still not sure.

eugenesvk:
> I'm trying the behavior in different terminals.
> Putty seems to have a setting where backspace always sends ^H.

But the default in putty is the second one, so it's not an issue? 
putty
Otherwise yeah, many terminals have only two codes (`8`/`0x08` and `127`/`0x7f`) for these three keys/combos
<kbd>Control</kbd><kbd>H</kbd>
<kbd>Control</kbd><kbd>Backspace</kbd>
<kbd>Backspace</kbd>
so whatever you set you won't get the full separation, need a terminal that supports a third key code `31`/`0x1f`


> What I'm also afraid of is that some users have the habit of literally typing control-h instead of backspace when they're in a terminal, because the key is easier to reach. So, I'm not sure we should make it a default on Linux. I'm still not sure.

Interesting that a two-hand/two-key combo is easier to reach than a one-hand/one-key BS, but whatever moves your fingers ;)
But yeah, I totally get that there isn't really "the ideal" option as some uses cases are mutually exclusive, I guess the default should reflect what's expected by most of your users (if anyone knows what that is :))

jonathanslenders:
>> Interesting that a two-hand/two-key combo is easier to reach than a one-hand/one-key BS, but whatever moves your fingers ;)

I'm personally using a pok3r keyboard where the arrow keys are under FN+jkli (where my fn is on caps lock). It's not unusual either to use ctrl-h as backspace.

eugenesvk:
> I'm personally using a pok3r keyboard where the arrow keys are under FN+jkli (where my fn is on caps lock). It's not unusual either to use ctrl-h as backspace.

Oh, great minds think alike :), I have exactly the same cursor bindings (also with a caps lock), though that serves as a supplemental cursor as I still like the dedicated one; and **caps lock+h** would make perfect sense in this setup, but **ctrl-H** is still a bit strange

eshepelyuk:
Is there any progress on this ?

carlwgeorge:
I generated a patch from this pull request (with a small adjustment to makes the tests pass), and I'm considering [applying it to the Fedora python-prompt-toolkit package](https://src.fedoraproject.org/rpms/python-prompt-toolkit/pull-request/6).  So far in my testing with ipython inside ptyxis, it works great and allows me to configure ipython to behave like a browser text field (backspace deletes the previous character, control+backspace deletes the previous word).

```
c = get_config()
c.TerminalInteractiveShell.shortcuts = [
    {
        'match_keys': ['escape', 'h'],
        'new_keys': ['c-h'],
        'command': 'prompt_toolkit:named_commands.backward_kill_word',
    },
]
```

PR #1401, opened 2021-03-19
Added in_thread parameter to `Application.run`.

jonathanslenders:
This is useful for running an application in a background thread, while the
main thread blocks. This way, we are sure not to interfere with an event loop
in the current thread.

### DIFF
--- a/prompt_toolkit/application/application.py
+++ b/prompt_toolkit/application/application.py
@@ -3,6 +3,7 @@ import os
 import re
 import signal
 import sys
+import threading
 import time
 from asyncio import (
     AbstractEventLoop,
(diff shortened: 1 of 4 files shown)


MrMino:
Cool! Thank you!

Note to self: might be useful for gotcha/ipdb#184 and ipython/ipython#12141?

jonathanslenders:
Once this is released, it would be good indeed to use this in IPython as well.

I just looked at the IPython code, which I wrote back then, and we're not closing the event loop that got implicitly created in the thread here. https://github.com/ipython/ipython/blob/master/IPython/terminal/debugger.py#L111

There's a good chance we will leak file descriptors (3 for every run) because of this.

jonathanslenders:
@MrMino : I think actually in the case of IPython too, it would be good to always run the prompt in another thread, not just for ipdb. I didn't check, but wonder whether otherwise IPython's ``embed()`` an an asyncio application works.

In ptpython, input is always read in a separate thread, and this works great now. Some notes related to this I wrote down, mainly for myself: https://github.com/prompt-toolkit/ptpython/blob/master/docs/concurrency-challenges.rst

PR #1414, opened 2021-04-09
Use os.get_terminal_size for vt100._get_size

coolreader18:
(https://docs.python.org/3/library/os.html#os.get_terminal_size)

Python's had the functionality implemented in the stdlib for a while, and this gets rid of any bugs related to mutability that the comment talks about (if that's still an issue `¯\_(ツ)_/¯`)


jonathanslenders:
Thanks!

PR #1422, opened 2021-04-30
make the vt100 flush method re-entrant

mguijarr:
I use prompt toolkit with gevent, thanks to the super-useful[ aiogevent library](https://github.com/2mf/aiogevent)

I managed to get two times identical output, this is happening when `flush()` is called
whereas it is still executing - the solution to this is to make the `flush()` method re-entrant, by
reinitializing `self._buffer` just after data have been read from it, instead of emptying it at the
end of the method.


jonathanslenders:
Thanks!

PR #1427, opened 2021-05-09
pyperclip: Fix pyperclip selection type

daniel-shimon:
Without this fix pasting single-line text always adds a newline


jonathanslenders:
Thanks!

PR #1429, opened 2021-05-13
Fix off-by-one error in FormattedTextControl mouse logic

djpohly:
Example: if the first fragment in a line is, say, `('', 'hello', handler)`, then `count` is 5, and `handler` should apply when `xpos` is 0, 1, 2, 3, or 4, but *not* 5, i.e. `count` should be strictly greater than `xpos`.

(To me it "feels right" to write this as `xpos < count` instead, since it's normal to compare an index to a length with less-than, but that may be a matter of personal taste, so I've left it as it was.)


jonathanslenders:
Thanks!

PR #1434, opened 2021-05-14
Fix OSError: [WinError 6] The handle is invalid on Windows, close  #1433

### LINKED ISSUE #1433: OSError: [WinError 6] The handle is invalid
Running on Windows 10

$ python -V
Python 3.8.4

$ ptpython -V
3.0.17

$ ptpython

print("ok")
Cancelling an overlapped future failed
future: <_OverlappedFuture pending overlapped=<pending, 0x235e16bedc0> cb=[BaseProactorEventLoop._loop_self_reading()]>
File "c:\users\lameg\appdata\local\programs\python\python38\lib\asyncio\windows_events.py", line 66, in _cancel_overlapped
self._ov.cancel()
OSError: [WinError 6] The handle is invalid
ok

joaompinto:


jonathanslenders:
I think it's this bug: https://bugs.python.org/issue39010

A fix was pushed to all Python releases having this bug: https://github.com/python/cpython/pull/22066

The ProactorLoop is considered better then the SelectorLoop, so we can probably do this, but only if it doesn't change the default for newer Python versions. I'd also prefer if it was not an import-time side effect. (but rather in a ``Application.run()`` or somewhere like that.)

PR #1446, opened 2021-06-18
Adds strikethrough text attribute.

### LINKED ISSUE #1416: Strikethrough text attribute
It looks that more and more (Linux) terminals support the strike through effect (it can be tested with `echo -e "\e[9mtest\e[0m"`), so I think, it would be great, if the `prompt-toolkit` can support this! Would that be feasible?

sphh:
- Adds tags: `<strike>`, `<s>` and `<del>`.
- Changes `pygments.generic.deleted` from "#a00000" to "strike".

**Note**
Only `html` and `vt100` is implemented.

Closes #1416.


prompt_toolkit/styles/defaults.py
```diff
@@ -193,7 +197,7 @@
     "pygments.literal.number": "#666666",
     "pygments.generic.heading": "bold #000080",
     "pygments.generic.subheading": "bold #800080",
-    "pygments.generic.deleted": "#a00000",
+    "pygments.generic.deleted": "strike",
```
jonathanslenders, line 200: This is the only change where I'm not sure whether it's a good idea. These class names are used by Pygments for rendering diffs. 1. underline is probably not everywhere that well supported. 2. people are used to red/green. So, I think I'll revert this line and choose the safe default.

jonathanslenders:
Looks mostly good. Thanks for this! Unit tests are still failing though.

prompt_toolkit/styles/defaults.py
sphh: I don't have any objections at all to change this back to `"a00000"`, if it is the better default!

sphh:
Oh, I never looked at the tests … It's done now and all tests are passing locally.

jonathanslenders:
Thanks!

prompt_toolkit/styles/defaults.py
sphh: @jonathanslenders: I notice that the commit https://github.com/prompt-toolkit/python-prompt-toolkit/commit/81fbd23172108584d47f8403b5954ec91971096c still has this line with `"strike"` instead of `"#a00000"`, because I have never changed it back, since I had the feeling that our discussion has not resolved this issue. Please feel free to change it back to `"#a00000"` …

PR #1455, opened 2021-07-07
Ensure that 'print_formatted_text' plays nice with patch_stdout.

jonathanslenders:
Before, it would ignore `patch_stdout` completely to begin with.
Now, it does not require it, but will always print above the prompt.


GNUGradyn:
This does not appear to have solved the problem. Please send example code showing this working

GNUGradyn:
Nevermind, just have to manually install from git instead of pypi

PR #1478, opened 2021-08-08
 Hack to fix non-bmp unicode input for windows

### LINKED ISSUE #1477: Exception on non-BMP character input in Windows Terminal
When attempting to input a non-BMP character, like 😀, in the (new) Windows Terminal, the following exception is thrown and the input fails:

```
Unhandled exception in event loop:
  File "c:\users\ryan\appdata\local\programs\python\python39\lib\asyncio\events.py", line 80, in _run
    self._context.run(self._callback, *self._args)
  File "c:\users\ryan\appdata\local\programs\python\python39\lib\site-packages\prompt_toolkit\input\win32.py", line 512, in ready
    callback()
  File "c:\users\ryan\appdata\local\programs\python\python39\lib\site-packages\prompt_toolkit\application\application.py", line 662, in read_from_input
    keys = self.input.read_keys()
  File "c:\users\ryan\appdata\local\programs\python\python39\lib\site-packages\prompt_toolkit\input\win32.py", line 75, in read_keys
    return list(self.console_input_reader.read())
  File "c:\users\ryan\appdata\local\programs\python\python39\lib\site-packages\prompt_toolkit\input\win32.py", line 233, in read
    all_keys = list(self._get_keys(read, input_records))
  File "c:\users\ryan\appdata\local\programs\python\python39\lib\site-packages\prompt_toolkit\input\win32.py", line 292, in _get_keys
    for key_press in self._event_to_key_presses(ev):
  File "c:\users\ryan\appdata\local\programs\python\python39\lib\site-packages\prompt_toolkit\input\win32.py", line 330, in _event_to_key_presses
    ascii_char = u_char.encode("utf-8")

Exception 'utf-8' codec can't encode character '\ud83d' in position 0: surrogates not allowed
```
It seems that the terminal is sending the character as separate utf-16 surrogate codes, which breaks the win32 input logic. Either the two key events would somehow have to be merged into one KeyPress with the full character or the unicode would have to be normalized somehow later on.

rmccampbell:
Fixes #1477. This transforms the list of decoded KeyPress events and merges surrogates into single Unicode characters.

### DIFF
--- a/prompt_toolkit/input/win32.py
+++ b/prompt_toolkit/input/win32.py
@@ -252,6 +252,9 @@ class ConsoleInputReader:
         # Fill in 'data' for key presses.
         all_keys = [self._insert_key_data(key) for key in all_keys]
 
+        # Correct non-bmp characters that are passed as separate surrogate codes
+        all_keys = list(self._merge_paired_surrogates(all_keys))
+
         if self.recognize_paste and self._is_paste(all_keys):
             gen = iter(all_keys)
             k: Optional[KeyPress]
@@ -316,6 +319,32 @@ class ConsoleInputReader:
                         yield key_press
 
     @staticmethod
+    def _merge_paired_surrogates(key_presses: List[KeyPress]) -> Iterator[KeyPress]:
+        buffered_high_surrogate = None
+        for key in key_presses:
+            is_text = not isinstance(key.key, Keys)
+            is_high_surrogate = is_text and '\uD800' <= key.key <= '\uDBFF'
+            is_low_surrogate = is_text and '\uDC00' <= key.key <= '\uDFFF'
+
+            if buffered_high_surrogate:
+                if is_low_surrogate:
+                    fullchar = ((buffered_high_surrogate.key + key.key)
+                        .encode('utf-16-le', 'surrogatepass')
+                        .decode('utf-16-le'))
+                    key = KeyPress(fullchar, fullchar)
+                else:
+                    yield buffered_high_surrogate
+                buffered_high_surrogate = None
+
+            if is_high_surrogate:
+                buffered_high_surrogate = key
+            else:
+                yield key
+
+        if buffered_high_surrogate:
+            yield buffered_high_surrogate
+
+    @staticmethod
     def _is_paste(keys: List[KeyPress]) -> bool:
         """
         Return `True` when we should consider this list of keys as a paste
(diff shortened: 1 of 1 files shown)


prompt_toolkit/input/win32.py
```diff
@@ -315,6 +318,37 @@ def _get_keys(
                     for key_press in self._handle_mouse(ev):
                         yield key_press
 
+    @staticmethod
+    def _merge_paired_surrogates(key_presses: List[KeyPress]) -> Iterator[KeyPress]:
```
jonathanslenders, line 322: Just wondering, is this similar to what ``unicodedata.normalize`` does, but for ``KeyPress`` objects?

prompt_toolkit/input/win32.py
```diff
@@ -260,7 +263,7 @@ def read(self) -> Iterable[KeyPress]:
                 # Pasting: if the current key consists of text or \n, turn it
                 # into a BracketedPaste.
                 data = []
-                while k and (isinstance(k.key, str) or k.key == Keys.ControlJ):
+                while k and (not isinstance(k.key, Keys) or k.key == Keys.ControlJ):
```
jonathanslenders, line 266: Is this change needed?

rmccampbell: This is unrelated to the main change but a potential bug fix. Because the `Keys` enum inherits `str`, technically the original check would always pass, so I think this change more correctly matches the intent of the code. Actually it would probably be better to make the enum not inherit `str` in the first place (which makes the enum a drop-in replacement for str in most cases); I don't think this is very useful but it may be relied on by some code so not sure if it would be safe to change without analyzing more of the code base.

prompt_toolkit/input/win32.py
rmccampbell: Not exactly, although a similar idea. unicodedata.normalize works at a higher level of abstraction, between "conceptual" characters and code points, i.e. it can merge 'a' and '◌̈' into 'ä'. This function is merging UTF-16 surrogate pair code units (see https://en.wikipedia.org/wiki/UTF-16) into single code points. Apparently the underlying windows API calls are working in UTF-16 and treating the code units as individual characters, when it ideally should return full code points, hence this workaround.

rmccampbell: Example: the character 😀 (U+1F600) can be encoded with the two 16-bit code units [0xD83D, 0xDE00].

jonathanslenders:
@rmccampbell: Thank you for this!
Merged from here: https://github.com/prompt-toolkit/python-prompt-toolkit/pull/1482 (squashed/reformatted).

PR #1485, opened 2021-08-23
Update to Sphinx 4.1.2

kiancross:
It looks as though the current Sphinx version is quite old. In particular, the [`objecst.inv`](https://python-prompt-toolkit.readthedocs.io/en/master/objects.inv) file currently generated for this project seems to be missing some objects.

In this PR, I have upgraded Sphinx to version 4.1.2. There did not appear to be any issues caused by doing this. This resolves the aforementioned problem with `objects.inv`.

I have put a test rendering of the documentation [here](https://prompt-toolkit-test.readthedocs.io/en/upgrade-sphinx/index.html), so you can also check that nothing has regressed.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,0 +1,14 @@
+version: 2
+
+formats:
+  - pdf
+  - epub
+
+sphinx:
+  configuration: docs/conf.py
+
+python:
+  install:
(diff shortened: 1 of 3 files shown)


jonathanslenders:
Thanks!

PR #1490, opened 2021-08-27
[Buffer] Support Buffer(document_class=[..]) to fix creation of Document instances (fixes #1484)

### LINKED ISSUE #1484: [Buffer] Creates new instances of plain Document that mis-match .document behavior when .document is an instance of an inherited class
Hi!

Thanks for making Prompt Toolkit — I use it happily in [git-delete-merged-branches](https://github.com/hartwork/git-delete-merged-branches) :pray: 

In `git-delete-merged-branches` my scenario is this: The user can configure the tool interactively, e.g. select one or more required merge target branches. Configuration shows up at the first start per repository, automatically. For example:

pt_1

That very prompt has interactive search enabled, e.g. as seen here:

pt_2

Now from a user perspective, I wanted (and got that working):
- To only find search matches in those branch name items but not the header text above or the help below.
- To have existing key binding <kbd>n</kbd> and <kbd>N</kbd> jump straight to the next/previous line, even if a branch name matches the search pattern multiple times (e.g. with branch `ref[a]ctor-l[a]yout-window` in the picture).

To achieve the line-jumping <kbd>n</kbd>, I have [derived a class `LineJumping Buffer`](https://github.com/hartwork/git-delete-merged-branches/blob/3cd8b5fa0aa31210058b4103b40252ad6dec10aa/git_delete_merged_branches/_multiselect.py#L72-L100) from `prompt_toolkit.buffer.Buffer` and works well.
However, for the part limiting the scope of the search to the list of the branches I have three parts where one seems troublesome:
- I have an input processor that [drops `class:search` and `class:incsearch`](https://github.com/hartwork/git-delete-merged-branches/blob/3cd8b5fa0aa31210058b4103b40252ad6dec10aa/git_delete_merged_branches/_multiselect.py#L54-L62) depending on line number, that works.
- I have derived a [class `ItemOnlySearchDocument`](https://github.com/hartwork/git-delete-merged-branches/blob/3cd8b5fa0aa31210058b4103b40252ad6dec10aa/git_delete_merged_branches/_multiselect.py#L278-L319) from `prompt_toolkit.document.Document` that overrides methods `find` and `find_backwards` to discard matches depending on line number, works well.
- Because `prompt_toolkit.buffer.Buffer` creates new instances of `prompt_toolkit.document.Document` internally, my overrides in `ItemOnlySearchDocument` where not applying everywhere. So (simplified) I wrapped everything by a rather hacky [`with patch('prompt_toolkit.buffer.Document', ItemOnlySearchDocument):`](https://github.com/hartwork/git-delete-merged-branches/blob/3cd8b5fa0aa31210058b4103b40252ad6dec10aa/git_delete_merged_branches/_multiselect.py#L329) to make `Buffer` instantiate my class, to then get the full effect.

That last point — the use of `unittest.mock.patch` — seems hacky enough, that I felt I should create this ticket and hear if you'd consider adding something like `document_class=Document` to the `Buffer` class, so that I no longer need to monkey patch things to get my inherited Document to apply everywhere during search.

What do you think?

Thanks and best, Sebastian

hartwork:
Hi! Here's _one way_ to fix #1484. Happy to adjust, let me know what you think. Thanks! :pray:

### DIFF
--- a/prompt_toolkit/buffer.py
+++ b/prompt_toolkit/buffer.py
@@ -23,6 +23,7 @@ from typing import (
     Optional,
     Set,
     Tuple,
+    Type,
     TypeVar,
     Union,
     cast,
@@ -250,6 +251,7 @@ class Buffer:
         on_cursor_position_changed: Optional[BufferEventHandler] = None,
         on_completions_changed: Optional[BufferEventHandler] = None,
         on_suggestion_set: Optional[BufferEventHandler] = None,
+        document_class: Type[Document] = Document,
     ):
 
         # Accept both filters and booleans as input.
@@ -298,7 +300,7 @@ class Buffer:
         # Document cache. (Avoid creating new Document instances.)
         self._document_cache: FastDictCache[
             Tuple[str, int, Optional[SelectionState]], Document
-        ] = FastDictCache(Document, size=10)
+        ] = FastDictCache(document_class, size=10)
 
         # Create completer / auto suggestion / validation coroutines.
         self._async_suggester = self._create_auto_suggest_coroutine()
@@ -309,6 +311,7 @@ class Buffer:
         self._load_history_task: Optional[asyncio.Future[None]] = None
 
         # Reset other attributes.
+        self.document_class = document_class
         self.reset(document=document)
 
     def __repr__(self) -> str:
@@ -328,7 +331,7 @@ class Buffer:
         if append_to_history:
             self.append_to_history()
 
-        document = document or Document()
+        document = document or self.document_class()
 
         self.__cursor_position = document.cursor_position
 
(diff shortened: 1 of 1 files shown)


hartwork:
Any thoughts?

hartwork:
Hi @jonathanslenders thanks for your review and consideration.

I have no hard feelings about this particular implementation but rather my goal is to have "any" implementation you would accept that does two things:
- allow altering the search behavior — my precise real-world example (with links to code) is described in issue #1484 —…
- in a way that works without monkey-patching Prompt Toolkit using `unittest.mock.patch`.

Regarding the `FastDictCache` dynamic type issue you mention, I could imagine that including the absolute class name string (including the package and module) as a fourth argument in the key tuple could be a potential workaround for the problem you mention: Preserving static typing plus protection against mixing instances of different classes.

What do you think?

hartwork:
@jonathanslenders any news?

hartwork:
> That sounds fine. I'm also not a fan of `unittest.mock.patch`. I'll see if I can get some time to think and work on this.

@jonathanslenders any news?

hartwork:
@jonathanslenders any news?

hartwork:
Closing after one year to keep my https://github.com/pulls page clean…

PR #1498, opened 2021-09-12
validate_while_typing seems to now default to True

tsvenson:
The default behavior seems to be True in 3.0.20, so I have updated the doc to reflect that.


jonathanslenders:
Thanks! I don't remember any change regarding this, so I guess it's been like that for some time now.

Issue #1440, opened 2021-06-06
deprecation warning with Python 3.10


jnoortheen:
Python 3.10 is still beta. This is just a heads up.  Here is the full warning

```
xsh-rVZlQJi1-py3.10/lib/python3.10/site-packages/prompt_toolkit/application/application.py:878: DeprecationWarning: There is no current event loop
  loop = get_event_loop()
```

jonathanslenders:
Thanks for bringing this to my attention!

Given that we are still supporting Python 3.6, I think we need a wrapper (in ``prompt_toolkit/eventloop/utils.py``) that uses ``get_running_loop`` if available and otherwise ``get_event_loop``. Is this something you're interested in contributing?

There are various places where it's used:
```python
./shortcuts/prompt.py
./shortcuts/dialogs.py
./shortcuts/progress_bar/base.py
./shortcuts/utils.py
./input/vt100.py
./input/win32.py
./history.py
./application/application.py
./contrib/telnet/server.py
./contrib/ssh/server.py
./eventloop/inputhook.py
./eventloop/async_generator.py
./eventloop/utils.py
```

jnoortheen:
@jonathanslenders yes I am. This means updating the github actions as well right?

mattmc3:
Any updates on this issue? A fresh install of Xonsh will emit this deprecation error on every command since Python 3.10 is now out.

Instead of adding something to `prompt_toolkit/eventloop/utils.py`, would it preferable to do a simple conditional import in the relevant files? I'm willing to attempt a PR:

```python
try:
    from asyncio import get_running_loop
except ImportError:
    from asyncio import get_event_loop as get_running_loop
```

mattmc3:
I opened a PR. See above.

henryiii:
FYI, brew has now switched the xonsh recipe to Python 3.10, so this pops up all the time on any brew installed xonsh. Brew has migrated over 150 formula, and is migrating the other 450, so this will pop up on anything that moves to Python 3.10 and uses prompt-toolkit.

mattmc3:
Vote to close this issue. Can confirm that prompt-toolkit 3.0.22 works great in Xonsh on Python 3.10 with my merged PR. For Xonsh users, you can simply `xpip install prompt-toolkit --upgrade` if you are on an earlier version and still seeing the deprecation warning.

PR #1505, opened 2021-10-06
Add wrapper function for get_event_loop

mattmc3:
- Project still supports Py3.6
- asyncio.get_event_loop is deprecated in Py3.10
- Add wrapper function in eventloop utils to account for this

All the `tox` tests pass, however as this is my first contribution to this project I would appreciate a double check that this is not a breaking change.

### DIFF
--- a/examples/ssh/asyncssh-server.py
+++ b/examples/ssh/asyncssh-server.py
@@ -10,6 +10,7 @@ from pygments.lexers.html import HtmlLexer
 
 from prompt_toolkit.completion import WordCompleter
 from prompt_toolkit.contrib.ssh import PromptToolkitSSHServer, PromptToolkitSSHSession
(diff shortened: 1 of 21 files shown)


mattmc3:
Also, forgot to mention - this PR fixes: #1440

mattmc3:
Some references for this issue:

- [Deprecation thread](https://bugs.python.org/issue39529)
- [CPython handling](https://github.com/python/cpython/blob/6c942a86a4fb4c8b731cb1bd2933dba554eb79cd/Lib/asyncio/events.py#L765)
- [Similar wrapper functionality](https://github.com/Azure/azure-uamqp-python/blob/9155d578f05cab41244bdc23a2aae4adc345c960/uamqp/utils.py#L20-L37)

PR #1512, opened 2021-10-19
Add support for some CSI 27 modified variants of Enter

RPigott:
xterm has 3 different disambiguated key modes intended to be used by applications that wish to distinguish keyboard modifiers for certain keypresses. These emit CSI 27 escapes with a modifiers bitmask field, and may be enabled by configuration.

We can't parse these properly, so just add support for a few common modified variants of Enter in the escape sequence list. These variants previously had no distinct encoding, so the CSI variants have been adopted by some other terms, namely foot.

### DIFF
--- a/prompt_toolkit/input/ansi_escape_sequences.py
+++ b/prompt_toolkit/input/ansi_escape_sequences.py
@@ -121,6 +121,14 @@ ANSI_SEQUENCES: Dict[str, Union[Keys, Tuple[Keys, ...]]] = {
     "\x1b[23;2~": Keys.F23,
     "\x1b[24;2~": Keys.F24,
     # --
+    # CSI 27 disambiguated modified "other" keys (xterm)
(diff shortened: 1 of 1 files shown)


jonathanslenders:
Thanks a lot!

I think at some point we can start passing modifiers to ``KeyPressEvent`` in the same way that we recently did for mouse events.

Having Control-Enter or Shift-Enter support is a big deal for many applications. (That way, we can have intuitive bindings for "force execute" and "force insert newline"). Do you know any other terminals supporting this?

RPigott:
There are a few competing formats for disambiguated keys. I wrote a little about it [here](https://codeberg.org/dnkl/foot/issues/628#issuecomment-271212).

I don't have a feature matrix or anything, so I'm not sure which terminals support what, but xterm has had this feature for ~20 years. I believe today emacs understands CSI u and neovim understands CSI 27.

jonathanslenders:
@RPigott: Great! Thanks for the information!

PR #1536, opened 2021-12-08
chore: add py3.10 to ci

jnoortheen:


jonathanslenders:
Thank you @jnoortheen !

PR #1538, opened 2021-12-12
Handle SIGINT (when sent from another process) and allow binding it.

jonathanslenders:


asmeurer:
I've been using my own SIGINT handler, which is just a thin wrapper around some cleanups + raise KeyboardInterrupt. I tried replacing it with this and it doesn't work. Does this only trigger when the prompt is active?

jonathanslenders:
This is indeed only active for the duration of the prompt. The request came from the Xonsh community. It appears that Readline listens for SIGINT too, and and external process can send a SIGINT to the process which behaves exactly like pressing control-c.

If you're setting up a custom SIGINT handler of your own, I wonder whether this change does not interfere with that and overrides that signal handler.

asmeurer:
I just realized my SIGINT handler is not necessary and I can handle KeyboardInterrupt directly. So this isn't really a problem for me like I thought it would be. 

I'm actually having a hard time figuring out if my SIGINT handler ever actually worked at all. I can't get it to trigger, even in older versions of prompt-toolkit. I was pretty sure I had it working, so I'm not sure what is different now.

PR #1542, opened 2022-01-09
Fix mypy errors

staticdev:
Current version has errors:

mypy prompt_toolkit:

```
prompt_toolkit/layout/utils.py:31: error: Argument 1 of "insert" is incompatible with supertype "list"; supertype defines the argument type as "SupportsIndex"
prompt_toolkit/layout/utils.py:31: note: This violates the Liskov substitution principle
prompt_toolkit/layout/utils.py:31: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#incompatible-overrides
prompt_toolkit/application/application.py:1165: error: Argument 2 to "kill" has incompatible type "Optional[Any]"; expected "int"
prompt_toolkit/application/application.py:1167: error: Argument 2 to "kill" has incompatible type "Optional[Any]"; expected "int"
Found 3 errors in 2 files (checked 143 source files)
```

### DIFF
--- a/prompt_toolkit/application/application.py
+++ b/prompt_toolkit/application/application.py
@@ -1156,15 +1156,16 @@ class Application(Generic[_AppResult]):
         if _SIGTSTP is not None:
 
             def run() -> None:
+                signal = cast(int, _SIGTSTP)
                 # Send `SIGTSTP` to own process.
                 # This will cause it to suspend.
 
                 # Usually we want the whole process group to be suspended. This
                 # handles the case when input is piped from another process.
                 if suspend_group:
-                    os.kill(0, _SIGTSTP)
+                    os.kill(0, signal)
                 else:
-                    os.kill(os.getpid(), _SIGTSTP)
+                    os.kill(os.getpid(), signal)
 
             run_in_terminal(run)
 
--- a/prompt_toolkit/layout/utils.py
+++ b/prompt_toolkit/layout/utils.py
@@ -1,4 +1,5 @@
 from typing import TYPE_CHECKING, Iterable, List, TypeVar, Union, cast, overload
+
 from typing_extensions import SupportsIndex
 
 from prompt_toolkit.formatted_text.base import OneStyleAndTextTuple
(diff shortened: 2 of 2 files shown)


prompt_toolkit/application/application.py
```diff
@@ -1156,15 +1156,16 @@ def suspend_to_background(self, suspend_group: bool = True) -> None:
         if _SIGTSTP is not None:
 
             def run() -> None:
+                signal = cast(int, _SIGTSTP)
```
jonathanslenders, line 1159: Thanks for the suggestions!

I'd probably prefer to cast immediately at the top of the file, if that works. Something like:
```python
_SIGTSTP = cast(Optional[int], getattr(signal, "SIGTSTP", None))
```

prompt_toolkit/layout/utils.py
```diff
@@ -1,5 +1,7 @@
 from typing import TYPE_CHECKING, Iterable, List, TypeVar, Union, cast, overload
 
+from typing_extensions import SupportsIndex
```
jonathanslenders, line 3: ``typing_extensions`` is not a required dependency, so this import should be an a ``if TYPE_CHECKING:`` block.

jonathanslenders (changes requested):
Thanks a lot! Two small remarks before I can merge.

prompt_toolkit/layout/utils.py
staticdev: I see, do you have plans to drop usage of python 3.6 soon as it is already End-Of-Life? With [pyupgrade](https://github.com/asottile/pyupgrade) one can easily migrate the whole code to use new style of annotations.

prompt_toolkit/application/application.py
staticdev: Unfortunately, that doesn't solve the error. Seems like mypy is not smart enough to filter it to int after `if _SIGTSTP is not None:`. The way I did was the only way I found that gets rid of this mypy error.

jonathanslenders:
Thanks!

PR #1568, opened 2022-02-10
merge html formatting updates into ansi

tapple:


prompt_toolkit/formatted_text/ansi.py
```diff
@@ -257,11 +258,17 @@ def format(self, *args: str, **kwargs: str) -> "ANSI":
         Like `str.format`, but make sure that the arguments are properly
         escaped. (No ANSI escapes can be injected.)
         """
-        # Escape all the arguments.
-        args = tuple(ansi_escape(a) for a in args)
-        kwargs = {k: ansi_escape(v) for k, v in kwargs.items()}
+        return ANSI(FORMATTER.vformat(self.value, args, kwargs))
 
-        return ANSI(self.value.format(*args, **kwargs))
+    def __mod__(self, value: Union[object, Tuple[object, ...]]) -> "ANSI":
```
jonathanslenders, line 263: The type for ``value`` can be ``object`` (A union with an ``object`` inside is always ``object``.)

tests/test_formatted_text.py
```diff
@@ -103,6 +103,66 @@ def test_ansi_true_color():
     ]
 
 
+def test_ansi_interpolation():
+    # %-style interpolation.
+    value = ANSI("\x1b[1m%s\x1b[0m") % "hello"
```
jonathanslenders, line 108: For the tests, it would be good to have at least a ``\x1b`` in the right hand argument, so that we can escaping happens properly.

jonathanslenders (changes requested):
Thanks for the PR! Two small changes would be required.

jonathanslenders (approved):

PR #1578, opened 2022-03-04
Add Github project link to documentation landing page

miohtama:
If a first time user lands on the documentation page, there is no obvious way to get to Github (though there are links).

Adding a link that says "Github here"


jonathanslenders:
Thanks!

Issue #1522, opened 2021-11-03
PosixPipeInput.close() hangs the application


wuf:
If pipe input is closed when the application (for example, prompt()) is running, the app is not aware of this event and it won't exit and cleanup.

This can be easily reproduced w/ the telnet service as it uses pipe to pass user input from socket to the prompt application, when user close the telnet connection in middle of the prompt(), the application coroutine will stay there forever.

The root cause is that both reading and writing fd are closed when closing the pipe, so the callback function of event loops reader won't be called. The correct way to give the callback a chance to read 0 bytes (to detect EOF) from the pipe reading end is to close the pipe's writing end only. (a similar issue & answer can be found at [stackoverflow](https://stackoverflow.com/questions/56367271/how-to-wait-on-a-error-close-event-for-a-socket-with-asyncio?noredirect=1&lq=1))

Below patch can be used to verify that prompt application can detect EOF if we only close the pipe's writing end when the socket connection is broken. This patch is a ugly hack for quick demo.

```diff
diff --git a/prompt_toolkit/contrib/telnet/server.py b/prompt_toolkit/contrib/telnet/server.py
index 9b325fd8..f039bcad 100644
--- a/prompt_toolkit/contrib/telnet/server.py
+++ b/prompt_toolkit/contrib/telnet/server.py
@@ -4,12 +4,13 @@ Telnet server.
 import asyncio
 import contextvars  # Requires Python3.7!
 import socket
+import os
+from asyncio import get_event_loop
 from typing import Awaitable, Callable, List, Optional, Set, TextIO, Tuple, cast

 from prompt_toolkit.application.current import create_app_session, get_app
 from prompt_toolkit.application.run_in_terminal import run_in_terminal
 from prompt_toolkit.data_structures import Size
-from prompt_toolkit.eventloop import get_event_loop
 from prompt_toolkit.formatted_text import AnyFormattedText, to_formatted_text
 from prompt_toolkit.input import create_pipe_input
 from prompt_toolkit.output.vt100 import Vt100_Output
@@ -132,6 +133,7 @@ class TelnetConnection:
         self._closed = False
         self._ready = asyncio.Event()
         self.vt100_output = None
+        self._partial_closed = False

         # Create "Output" object.
         self.size = Size(rows=40, columns=79)
@@ -180,7 +182,10 @@ class TelnetConnection:
             else:
                 # Connection closed by client.
                 logger.info("Connection closed by client. %r %r" % self.addr)
[...]

PR #1585, opened 2022-03-07
bugfix telnet server: prompt app not cancelled when client close connection

wuf:
If client closes the telnet connection (either due to user typing
   "Ctrl-] then 'quit'" or due to network shutdown) in middle of the
   prompt application, the application will hang forever and resources
   maintained in TelnetServer.connections and
   TelnetServer._application_tasks are not released.

### DIFF
--- a/prompt_toolkit/contrib/telnet/server.py
+++ b/prompt_toolkit/contrib/telnet/server.py
@@ -87,6 +87,7 @@ class _ConnectionStdout:
         self._connection = connection
         self._errors = "strict"
         self._buffer: List[bytes] = []
+        self._closed = False
 
     def write(self, data: str) -> None:
         data = data.replace("\n", "\r\n")
@@ -98,11 +99,15 @@ class _ConnectionStdout:
 
     def flush(self) -> None:
         try:
-            self._connection.send(b"".join(self._buffer))
+            if not self._closed:
+                self._connection.send(b"".join(self._buffer))
         except OSError as e:
             logger.warning("Couldn't send data over socket: %s" % e)
 
         self._buffer = []
+
+    def close(self) -> None:
+        self._closed = True
 
     @property
     def encoding(self) -> str:
@@ -137,6 +142,7 @@ class TelnetConnection:
         self._closed = False
         self._ready = asyncio.Event()
         self.vt100_output = None
+        self._task: Any = None
 
         # Create "Output" object.
         self.size = Size(rows=40, columns=79)
@@ -186,6 +192,9 @@ class TelnetConnection:
                 # Connection closed by client.
                 logger.info("Connection closed by client. %r %r" % self.addr)
                 self.close()
+                # to abort the prompt app
+                if self._task:
+                    self._task.cancel()
 
         # Add reader.
         loop = get_event_loop()
@@ -196,7 +205,10 @@ class TelnetConnection:
             await self._ready.wait()
             with create_app_session(input=self.vt100_input, output=self.vt100_output):
                 self.context = contextvars.copy_context()
-                await self.interact(self)
+                self._task = get_event_loop().create_task(self.interact(self))
+                if self._task:
+                    await self._task
+                    self._task = None
         except Exception as e:
             print("Got %s" % type(e).__name__, e)
             import traceback
@@ -222,6 +234,7 @@ class TelnetConnection:
             self.vt100_input.close()
             get_event_loop().remove_reader(self.conn)
             self.conn.close()
+            self.stdout.close()
 
     def send(self, formatted_text: AnyFormattedText) -> None:
         """


jonathanslenders:
Hi @wuf,

Thanks for reporting the issue and proposing this patch!

The underlying issue is more complex, and it looks like there are a couple of bugs. Some things here that I notice:

- I noticed that some of the telnet-related examples in the examples/ folder are not up to date and don't run as-is. This should be simple to fix.
- The more I learn about structured concurrency, the more I think it's the right approach for cancellation in this case. Anything spawned as part of a connection should be spawned as part of a task group, for which the lifetime doesn't extend the lifetime of the connection. For now, however, I don't want to depend on anyio or have a task group implementation in prompt_toolkit, but I regret having a separate ``.start()`` and ``.stop()`` method in ``TelnetServer`` instead of a single ``.run()`` coroutine will be stopped through cancellation. It's a breaking change that I hope to fix it at some point.
- Cancelling the interact task at the end of the input is not necessarily the right thing to do. The application should be allowed to send some final output after the input terminates. So, instead, we want to raise ``EOFError`` when more input is asked for if the input is closed (this doesn't happen because of the next point).
- It looks like there is a bug in ``PosixPipeInput``. The ``.close()`` method closes both the read and write end of the pipe. Closing the read-end causes the attached callback in ``.vt100.py -> _attached_input()`` to never wake up. Instead, we should only close the write-end, after which the loop wakes up, calls the callback for the read-end of the pipe where we see that the input is closed, and ``EOFError`` will be raised because of that. However, not closing the read-end of the pipe means that it has to be closed elsewhere, so ``PosixPipeInput`` should probably become a context manager. That is a breaking change.
- In ``telnet/server.py``, in the ``run_application()`` method, it looks like exceptions that are caught, are reraised. This includes an ``EOFError`` that was raised in the interact coroutine. We should probably not be reraising exceptions here.

I hope this is sufficient as feedback to set the direction. I can't merge it as-is unfortunately. I'll see whether I can already take care of some bullet points here.
By the way, thanks for the ``"Ctrl-] then 'quit'" `` trick. I didn't know about that. This is very useful for testing.

jonathanslenders:
Hi @wuf,

This is my attempt to solve the issue related to closing the connection: https://github.com/prompt-toolkit/python-prompt-toolkit/pull/1586
(It does not deal with all issues yet, and still needs some testing.)

wuf:
Hi @jonathanslenders,

thanks for the quick reply. Yes, I totally agree with your points and this should be fixed at PosixPipeInput.  I once raised the PosixPipe issue [#1522](url) to report this but I can only provide an ugly demo at that time.

Gald that this issue is finally noticed and a mature fix is provided by you. I will try it in my application, thanks.

wuf:
@jonathanslenders ,

Your patch works, below are a few things that I noticed,

- how about adding `__enter__()` and `__exit__()` to turn PosixPipeInput instances into context manager? Then the changes would be compatible to existing code.
- another issue is that when pipe is closed, renderer still tries to write/flush stdout which causes several below warning messages. In my previous patch, I added a `_closed` flag to `_ConnectionStdout` to address this. Or maybe we can just lower the logging level?

`WARNING:prompt_toolkit.contrib.telnet:Couldn't send data over socket: [Errno 9] Bad file descriptor`

jonathanslenders:
@wuf: Thanks for the very quick reply!

- About ``__enter__`` and ``__exit__``. In general, I started to avoid those for two reasons. First, they don't require that the object will be used as a context manager. Secondly, state (local variables) is often passed from the ``__enter__`` to the ``__exit__`` phase. It doesn't really apply here, but this experience and what I learned about structured concurrency, makes me only want to write context managers anymore using ``contextlib.contextmanager`` and ``contextlib.asynccontextmanager``. You have a point about backwards compatibility, but it's hard to guarante that these objects will remain usable without context manager. They will leak file descriptors in any case.

- About the second issue. I just fixed that, exactly like you did.

jonathanslenders:
@wuf: I merged my branch (I made you co-author in the commit message). Feel free to still comment, if there is anything.

wuf:
@jonathanslenders,

- the `_closed` flag wasn't taking effects
- it feels that the EOFError traceback is little bit annoying:  closing telnet connection is a normal user operation, and we already has the printing "Connection closed by client" for it.

how about below patch

``` diff
diff --git a/prompt_toolkit/contrib/telnet/server.py b/prompt_toolkit/contrib/telnet/server.py
index 035239d2..227fc1fe 100644
--- a/prompt_toolkit/contrib/telnet/server.py
+++ b/prompt_toolkit/contrib/telnet/server.py
@@ -99,7 +99,8 @@ class _ConnectionStdout:

     def flush(self) -> None:
         try:
-            self._connection.send(b"".join(self._buffer))
+            if not self._closed:
+                self._connection.send(b"".join(self._buffer))
         except OSError as e:
             logger.warning("Couldn't send data over socket: %s" % e)

@@ -355,6 +356,9 @@ class TelnetServer:
                     finally:
                         self.connections.remove(connection)
                         logger.info("Stopping interaction %r %r", *addr)
+            except EOFError:
+                # caused by "Connection closed by client"
+                pass
             except BaseException as e:
                 print("Got %s" % type(e).__name__, e)
                 import traceback

```

jonathanslenders:
I'm closing this. Thanks again @wuf!

PR #1590, opened 2022-03-10
Fix some small list -> Iterable/Sequence typing issues

CarliJoy:
"List" are invariant (see https://mypy.readthedocs.io/en/stable/common_issues.html#variance), so we really should try to avoid them, especially as input parameter for our functions.

Stubled upon the following places where it clashed with my type checking.

Maybe there are more but start small ;-)

### DIFF
--- a/prompt_toolkit/formatted_text/base.py
+++ b/prompt_toolkit/formatted_text/base.py
@@ -44,7 +44,7 @@ if TYPE_CHECKING:
 AnyFormattedText = Union[
     str,
     "MagicFormattedText",
-    StyleAndTextTuples,
+    StyleAndTextTuplesTypes,
     # Callable[[], 'AnyFormattedText']  # Recursive definition not supported by mypy.
     Callable[[], Any],
     None,


prompt_toolkit/formatted_text/base.py
```diff
@@ -21,9 +22,11 @@
     Tuple[str, str], Tuple[str, str, Callable[[MouseEvent], None]]
 ]
 
-# List of (style, text) tuples.
+# List (covariant version of list) of (style, text) tuples
+# Not there is a subclass of this 
 StyleAndTextTuples = List[OneStyleAndTextTuple]
-
+# For typing stuff, we prefer an Iterable as it is more flexible
+StyleAndTextTuplesType = Iterable[OneStyleAndTextTuple]
```
jonathanslenders, line 29: I'm not sure that ``Iterable`` is desirable, because it means it's not indexable and possibly, can only be consumed once.
Without reading again through all of the prompt_toolkit codebase, I think ``Sequence`` is at least what we need.

prompt_toolkit/completion/word_completer.py
```diff
@@ -31,7 +31,7 @@ class WordCompleter(Completer):
 
     def __init__(
         self,
-        words: Union[List[str], Callable[[], List[str]]],
+        words: Union[Iterable[str], Callable[[], Iterable[str]]],
```
jonathanslenders, line 34: Please make this at least a ``Sequence``.

Otherwise, users are going to pass in a generator of strings, and wonder why the second time, they're not getting any completions.

Is there a reason it should be ``Iterable`` here?

CarliJoy: Thanks for the review @jonathanslenders 

Nope. I was just going to use the lowest possible. You are right a `Sequence` would be more fitting.

I stopped working on this PR as some  other troubles occured (see failing tests)
Once I found time to resolve them, I will undraft the PR and ping you.

PR #1602, opened 2022-03-24
Allow zero-width-escape sequences in `print_formatted_text`

joouha:
Hello,

This PR enables `[ZeroWidthEscape]` fragments to be sent as raw output in `print_formatted_text`.

---

@jonathanslenders I'm a bit confused about what you meant by the comment:

>     # Assume that the output is raw

because you're using `output.write` here, so escape codes are removed. Was your intention to use `output.write_raw` for everything here?

Thanks

### DIFF
--- a/src/prompt_toolkit/renderer.py
+++ b/src/prompt_toolkit/renderer.py
@@ -799,12 +799,16 @@ def print_formatted_text(
                 output.reset_attributes()
         last_attrs = attrs
 
-        # Eliminate carriage returns
-        text = text.replace("\r", "")
-
-        # Assume that the output is raw, and insert a carriage return before
-        # every newline. (Also important when the front-end is a telnet client.)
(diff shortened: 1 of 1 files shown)


jonathanslenders:
Hi @joouha,

My comment there is confusing. With "raw", I mean that the pseudo terminal is configured in raw mode, not cooked or in cbreak mode. (see: https://en.wikipedia.org/wiki/Terminal_mode ).

The method ``write_raw``, refers to a different "raw", where "raw" means don't remove/escape vt100 ANSI escape sequences.

jonathanslenders:
Thanks! I merged it. This makes sense.

Do you have an example where this is used? (Just wondering.)

joouha:
Thanks!

I wanted this for my `FormattedText` markdown formatter, to be able to display OSC 8 hyperlinks properly:

image

PR #1612, opened 2022-04-04
Add __bool__ to auto_suggest.Suggestion

MrMino:
This caught us by surprise in ipython/ipython#13624. It's easy to assume that `event.current_buffer.suggestion` is always either `None` or a non-zero length suggestion, but it turns out that it can actually be empty. This happens after accepting one suggestion and adding its text to the input. Afterwards, all subsequent events will have a `Suggestion` object with `self.text == ""`.

### DIFF
--- a/src/prompt_toolkit/auto_suggest.py
+++ b/src/prompt_toolkit/auto_suggest.py
@@ -46,6 +46,9 @@ class Suggestion:
     def __repr__(self) -> str:
         return "Suggestion(%s)" % self.text
 
+    def __bool__(self) -> bool:
+        return bool(self.text)
+
 
 class AutoSuggest(metaclass=ABCMeta):
     """


jonathanslenders:
Hi @MrMino,

Can you add a comment (not docstring) to the ``bool`` method referencing this issue and/or the IPython issue? That way it's clear why this helps.

In general, I'm not in favor of adding ``__bool__`` to random classes because the meaning is often very ambiguous (what if a class evolves and more fields are added, should they be included in the truthness-check or not?). Mypy's type narrowing encourages a style where you use ``if object is not None`` rather than ``if object:``. There is also this configuration option that should have prevented the IPython error: https://mypy.readthedocs.io/en/stable/error_code_list2.html?highlight=__bool__#check-that-expression-is-not-implicitly-true-in-boolean-context-truthy-bool

Anyway, I don't mind merging this, if it helps, if you add a comment as to why. Thanks!

MrMino:
I added it without thinking about it too much, that is a valid reasoning and to be honest now I feel kinda opposed to merging this.

I should probably just get my mypy setup in working order, it's been degraded for some time now. I now see that it's my workspace that is the issue, not the lack of `Suggestion.__bool__`.

PR #1563, opened 2022-02-07
add default values option for Dialog

motty-mio2:
Hello, I'm happy to post here.

## Why
Default Value settings are important for some people (includes me).

## What
I add an option to set default value for Dialogs(Checkbox list dialog and Radio list dialog).

- It can check whether the value is valid or invalid.
  - later checkbox example, "banana" is not a valid option, so it is removed. "eggs" and "croissants" are selected.

### Checkbox list dialog
```
results_array = checkboxlist_dialog(
    title="CheckboxList dialog",
    text="What would you like in your breakfast ?",
    values=[
        ("eggs", "Eggs"),
        ("bacon", "Bacon"),
        ("croissants", "20 Croissants"),
        ("daily", "The breakfast of the day")
    ],
    default_values=["banana", "eggs", "croissants"]
).run()
All values in default_values will be used for default value.
```
### Radio list dialog
```
result = radiolist_dialog(
    values=[
        ("red", "Red"),
        ("green", "Green"),
        ("blue", "Blue"),
        ("orange", "Orange"),
    ],
    default_values=["blue"],
    title="Radiolist dialog example",
    text="Please select a color:",
).run()
```
First index (default_values[0], in this case "blue") item will be used for default value.

### checkbox dialogs
I set default_values as `["banana", "eggs", "croissants"]`
checkbox
  
### radio dialogs
I set default_values as `["blue"]`
"blue" is selected.  
radio

## Related
- issue
    - #1544 
- PR
    - #651

Thank you.

Issue #1544, opened 2022-01-17
[Question] How to pass a default value to input_dialog()


martinburchell:
I'm looking for a way to pass a default value to `input_dialog()`. Is there one? From the code it doesn't look like it.

It looks like the default value would just need to be passed as the `text` parameter to TextArea() (not to be confused with the `text` parameter to `input_dialog()`).

If this is the case, would you accept a PR or would you rather keep `input_dialog()` simple? I note there is a related PR https://github.com/prompt-toolkit/python-prompt-toolkit/pull/651

PR #1616, opened 2022-04-14
add default value option for input dialog

motty-mio2:
Hello
I like this library so much and use it for me personal projects thanks. This PR is similar to previous (#1563).

# What
I add an option to set default value for input dialog

# Usage
```python
from prompt_toolkit.shortcuts import input_dialog


def main():
    result = input_dialog(
                title="Input dialog example", 
                text="Please type your name:", 
                default_text="George"
            ).run()

    print(f"Result = {result}")


if __name__ == "__main__":
    main()
```

# How it works
- without edit
default_value

- editing
edit

# Related
- issue
  -  #1544 
- PR
  - #1563 (me)

### DIFF
--- a/src/prompt_toolkit/shortcuts/dialogs.py
+++ b/src/prompt_toolkit/shortcuts/dialogs.py
@@ -109,7 +109,7 @@ def input_dialog(
     validator: Optional[Validator] = None,
     password: FilterOrBool = False,
     style: Optional[BaseStyle] = None,
-    default_text: str = "",
+    default: str = "",
 ) -> Application[str]:
     """
     Display a text input box.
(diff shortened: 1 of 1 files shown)


src/prompt_toolkit/shortcuts/dialogs.py
```diff
@@ -109,6 +109,7 @@ def input_dialog(
     validator: Optional[Validator] = None,
     password: FilterOrBool = False,
     style: Optional[BaseStyle] = None,
+    default_text: str = "",
```
jonathanslenders, line 112: Can you rename the parameter to ``default`` here?
For consistency with ``PromptSession``: https://github.com/prompt-toolkit/python-prompt-toolkit/blob/master/src/prompt_toolkit/shortcuts/prompt.py#L899

I know for the ``radiolist_dialog``, it's ``default_values``, but I think there it can be more than one.

motty-mio2: Ok

jonathanslenders:
Thanks!

PR #1617, opened 2022-04-19
Add has_suggestion filter

bitkeen:
In my ipython configuration, I add Ctrl-a and Ctrl-e bindings for beginning and end of line in vi mode.
Adding this filter would allow to map Ctrl-e both for end-of-line and for accepting suggestions.


jonathanslenders:
Thanks!

PR #1629, opened 2022-05-07
Add documentation for WindowRenderInfo, Point, Size

MrMino:
These classes are mentioned a few times in the reference but their documentation wasn't visible.
`Point` and `Size` might use some improvements, I added them mostly to make it possible to reference them properly.

image

image


jonathanslenders:
Thanks!

PR #1630, opened 2022-05-08
Add additional text styles (dim, overline, double underline, curvy underline, blink, & blink fast)

joouha:
This PR implements additional text styles:

- Dim
- Overline
- Double underline
- Curvy underline
- Blink
- Blink fast

<details>
<summary>Screenshot</summary>

extra-styles webp

</details>


jonathanslenders:
Thanks!

Before merging this it is also important to benchmark the performance impact on various platforms. Both CPU and memory. This data structure is initialized *many* times, and it would be a shame to merge this if it significantly impacts the performance on various systems, for a feature that's only available in a few terminals.

ryran:
Man I reeeeally want this... "dim" support especially. Sigh. Would this make `ANSI()` support the corresponding raw ansi codes (like dim)? Or would that need to be a separate bit of work altogether?

PR #1645, opened 2022-06-17
Add git-bbb to PROJECTS.rst

MrMino:
Something I've been working on for last month or two. It was super fun to build, and it proved useful when trying to understand large codebases with a lot of history (*\*cough\* IPython \*cough\**).


jonathanslenders:
Nice!

PR #1646, opened 2022-06-20
Update terminal suggestion for Windows

ofek:
Noticed its absence while passing by


jonathanslenders:
Actually, this comment is outdated. Now, both Windows terminal and cmd.exe are both fine, since we got PTY support in Windows. No need to use cmder/conemu.

PR #1656, opened 2022-07-26
Correct type hint for `OneStyleAndTextTuple` callable

joouha:
Hi,

I think the type of a `StyleAndTextTuple` should allow `NotImplemented` to be returned by the mouse handler, as is the case for other mouse handlers.

### DIFF
--- a/src/prompt_toolkit/formatted_text/base.py
+++ b/src/prompt_toolkit/formatted_text/base.py
@@ -5,6 +5,8 @@ from prompt_toolkit.mouse_events import MouseEvent
 if TYPE_CHECKING:
     from typing_extensions import Protocol
 
+    from prompt_toolkit.key_binding.key_bindings import NotImplementedOrNone
+
 __all__ = [
     "OneStyleAndTextTuple",
     "StyleAndTextTuples",
@@ -18,7 +20,7 @@ __all__ = [
 ]
 
 OneStyleAndTextTuple = Union[
-    Tuple[str, str], Tuple[str, str, Callable[[MouseEvent], None]]
+    Tuple[str, str], Tuple[str, str, Callable[[MouseEvent], "NotImplementedOrNone"]]
 ]
 
 # List of (style, text) tuples.


jonathanslenders:
Hi @joouha, can you elaborate? Do you have some code example that doesn't typecheck with the current annotations. I'm not sure I understand why.

joouha:
The problem is that the type annotations imply that a mouse handler must always return `None`.

This means that a mouse handler in a `FormattedTextControl` would invalidate the app after every mouse event.

So if you create a `FormattedTextControl` which responds to mouse-click events and not mouse-over events, the app still gets invalidated on mouse-over events. (This was causing me performance issues).

To prevent the mouse handler responding to mouse-over events, you need to return `NotImplemented`, but as typed currently, mouse handlers are only allowed to return `None`.

Here is a simple example app which demonstrates the problem:

image


<details>
<summary><strong>Expand for source code</strong></summary>

```python
import typing

from prompt_toolkit.application import Application, get_app
from prompt_toolkit.key_binding.key_processor import KeyPressEvent
from prompt_toolkit.layout import (
    FormattedTextControl,
    Layout,
    VSplit,
    Window,
    WindowAlign,
)
from prompt_toolkit.mouse_events import MouseEvent, MouseEventType
from prompt_toolkit.widgets import Box, Frame

if typing.TYPE_CHECKING:
    from prompt_toolkit.key_binding.key_bindings import NotImplementedOrNone

N_CLICKS = 0


def custom_mosue_handler(mouse_event: MouseEvent) -> "NotImplementedOrNone":
    """Update clicks, do not invalidate the app on other mouse events."""
    global N_CLICKS

    if mouse_event.event_type == MouseEventType.MOUSE_DOWN:
        N_CLICKS += 1
        return None
    else:
        # return None
        return NotImplemented


container = VSplit(
    [
        Box(
            Frame(
                Window(
                    FormattedTextControl(lambda: f"\n{get_app().render_counter}\n"),
                    align=WindowAlign.CENTER,
                ),
                title="Render Count",
                width=20,
                height=5,
            ),
            padding=0,
        ),
        Box(
            Frame(
                Window(
                    FormattedTextControl(lambda: f"\n{N_CLICKS}\n"),
                    align=WindowAlign.CENTER,
                ),
                title="Click Count",
                width=20,
                height=5,
            ),
            padding=0,
        ),
        Frame(
            Window(
                FormattedTextControl(
                    [("", "\nClick Me", custom_mosue_handler)],
                ),
                align=WindowAlign.CENTER,
                width=20,
                height=3,
                style="white bg:red",
            ),
        ),
    ]
)

kb = KeyBindings()


@kb.add("q")
def _quit(e: KeyPressEvent) -> None:
    """Exit the application."""
    e.app.exit()


app: Application[None] = Application(
    layout=Layout(container), key_bindings=kb, mouse_support=True, full_screen=True
)
app.run()
```

</details>

I do not want the app to invalidate when the mouse moves over the red control (the render count should not increase), so I return `NotImplemented` in the mouse handler for anything except a mouse-click. 

However, this fails with `mypy`, because mouse handlers can only return `None`:

```bash
❯ mypy examples/example.py

examples/example.py:63: error: List item 0 has incompatible type "Tuple[str, str, Callable[[MouseEvent], object]]"; expected "Union[Tuple[str, str], Tuple[str, str, Callable[[MouseEvent], None]]]"
Found 1 error in 1 file (checked 1 source file)
```

joouha:
Just to add:

- The return type of `FormattedTextControl.mouse_handler` is `NotImplementedOrNone`
- `FormattedTextControl.mouse_handler` returns the result of the calling the third element of a `OneStyleAndTextTuple`
- This suggests that the third element of a `OneStyleAndTextTuple` can validly return `NotImplemented`

jonathanslenders:
This makes sense, thanks!

PR #1663, opened 2022-08-26
Prevent `erase_when_done` affecting full screen apps

joouha:
If `erase_when_done` and `full_screen` are both set to `True` for an application, then the renderer never gets reset. This causes issues such as the cursor does not getting shown if it is hidden when exiting an application.

This PR works around this by ignoring `erase_when_done` for full-screen apps. 

I think `erase_when_done` should not really apply to full screen apps, as they effectively will always get erased when exiting the alternate screen.

I think it's unlikely that anyone depends on the existing behaviour, so hopefully it wouldn't break anything for anyone - let me know what you think.


jonathanslenders:
I think we should call `output.show_cursor()` in `Renderer.erase()`, similar to to the way we reset styling attributes and autowrap.

PR #1681, opened 2022-10-19
remove unused variable and performance optimization

vicalloy:
generator is faster than list.
```
import time

s = time.time()
max((e for e in range(10000*10000)), default=0)
print(time.time() - s)  # 3.0133349895477295

s = time.time()
max([e for e in range(10000*10000)], default=0)
print(time.time() - s)  # 4.363456964492798
```


jonathanslenders:
Thank you!

PR #1692, opened 2022-12-03
Create background tasks on application's loop

joouha:
I'm trying to load parts of my application (including a `Buffer` with history) in a background thread. However, when the background tasks are waited for at exit, I get the following error:

```
ValueError: The future belongs to a different loop than the one specified as the loop argument
```

The reason for this is that the buffer history loading task gets scheduled in the event loop in the background thread instead of in the main application loop.

This PR ensures that an application's background tasks always get created on the application's event loop (if it exists).

### DIFF
--- a/src/prompt_toolkit/application/application.py
+++ b/src/prompt_toolkit/application/application.py
@@ -1077,7 +1077,8 @@ class Application(Generic[_AppResult]):
 
         This is not threadsafe.
         """
-        task: asyncio.Task[None] = get_event_loop().create_task(coroutine)
+        loop = self.loop or get_event_loop()
+        task: asyncio.Task[None] = loop.create_task(coroutine)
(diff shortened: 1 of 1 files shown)


jonathanslenders:
Thanks!

Merging this, although I'm not 100% sure whether I understand the issue right.

I'm planning at some point to transition `Application.run()` to adopt `asyncio.run()` rather than `loop.run_until_complete()`. That means that when an application is started with `Application.run()`, it's always started in a fresh event loop, and that loop will be terminated when the application stops. It can't be reused for the next `Application.run()`. If this is not desirable, `Application.run_async` should be used. Also, at some point in the future, a `TaskGroup` will be used for these background tasks (in a couple of years from now, when we can drop Python 3.10 support ;) ). Just wanted to point that out in case that would cause any problems in the future.

PR #1693, opened 2022-12-04
Copy over show_cursor attribute from screen

rhlee:
If a ScrollablePane instance does not have focus, it will show the cursor of the currently focused window even if that window has show_cursor set to False. This is because the virtual screen does not copy over the show_cursor attribute from screen to the virtual screen.


jonathanslenders:
Thanks!

PR #1697, opened 2022-12-10
Fix currentThread() deprecation warning

smuth4:
`DeprecationWarning: currentThread() is deprecated, use current_thread() instead`

No change in functionality, just noticed the warning while testing a third-party package.


jonathanslenders:
Thanks!

PR #1701, opened 2022-12-17
Allow nested scrolling of Windows

joouha:
These changes mean that Windows will return `NotImplemented` when attempting to scroll beyond their scrollable limits.

This means mouse scroll events can bubble up to a scrollable parent container when a window has finished scrolling.

In the following screen recording, you can see that when the `TextArea` is moves under the mouse while scrolling the parent pane, it begins scrolling, and when the `TextArea` has scrolled all the way to the bottom, the parent pane resumes scrolling:

[nested-scrolling.webm](https://user-images.githubusercontent.com/12154190/208240041-c033d127-6a69-40a0-ac3a-a9059726d3a5.webm)

I've also changed the reported line-count of the `DummyControl` to `1` rather than `100*100` for this to work. Without this, Windows without content believe their `DummyControl` is always scrollable and accept all scroll events. I can't find any issues with doing this - why was the `DummyControl`'s content's line count set to `100*100`?


joouha:
I don't think the test failures are related to the proposed changes

joouha:
I've rebased and fixed the issues with checks.

This would also allow implementing a wrapping `Container` which can automatically disable mouse support for an app if you've scrolled to the top, allowing the terminal scrollback buffer to be scrolled instead.

[Screencast_20240117_193406-1.webm](https://github.com/prompt-toolkit/python-prompt-toolkit/assets/12154190/5676c390-6b5f-4212-8534-71601b6c0aa2)

PR #1708, opened 2023-01-17
Make VERSION tuple numeric.

### LINKED ISSUE #1707: Any reason why VERSION is a tuple of str and not int ?
AFACT it is a foot-gun as the tuple will be compared lexicographically (is `'36' < '4'`) is true.

I can imagine there might be legitimate reasons, but I could't find any looking a the codebase.

Carreau:
Note that by being numeric we drop the rc/a/b/dev suffix. I'm also adding the pep440 regular expression to make sure the version number match the spec.

The pep 440 spec is a bit annoying with the fact that a/b/rc don't have dots, but .dev do.

There are some issues if you don't follow the spec and publish both whl/tgz, as pip will mix-up the values, and think the tgz are actually a more recent version than the whl. It may be fixed in setuptools/pip, but better be safe.

Ii did not go the route of setting VERSION and having __version__ be a '.'.join() as setup.py use regex to extract version number.

Maye docs/conf.py could also use the same trick as setup.py to avoid having to update the version number in multiple places ?

Closes #1707

### DIFF
--- a/src/prompt_toolkit/__init__.py
+++ b/src/prompt_toolkit/__init__.py
@@ -13,6 +13,13 @@ See the examples directory to learn about the usage.
 Probably, to get started, you might also want to have a look at
 `prompt_toolkit.shortcuts.prompt`.
 """
+import re
+
+# note: this is a bit more lax than the actual pep 440 to allow for a/b/rc/dev without a number
+pep440 = re.compile(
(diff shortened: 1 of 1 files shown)


jonathanslenders:
Hi @Carreau , I see that you're assigning to the `pep440` variable, but I don't see it used anywhere. Is there anything that I'm missing?

BTW: about the failing tests, I just fixed that: https://github.com/prompt-toolkit/python-prompt-toolkit/pull/1711

Carreau:
Yeah, it's better to add before committing....

Rebased/squashed and force pushed. 
I've relaxed also the regex a bit compared to the official one that is meant to work only for releases.

jonathanslenders:
Thank you!

PR #1709, opened 2023-01-17
Documentation: Modernize Sphinx and update `autodoc` references

amotl:
Dear Jonathan,

first things first: Thanks a stack for conceiving and maintaining this excellent package, and a happy new year.

After discovering GH-1421, we thought it would be a reasonable idea to submit a patch which gives the documentation some love, and we hope you will like it. We have been able to make it to a point where only 7 warnings are issued when building the documentation.

With kind regards,
Andreas.

PR #1710, opened 2023-01-17
Fix mypy configuration by ignoring two files, to satisfy CI

amotl:
Hi again,

at GH-1709, we discovered that the test suite on CI currently croaks from an error with mypy. mypy raised this error on four occasions (see below):
```
error: "_SupportsAcloseT" has no attribute "__aiter__" (not async iterable)  [attr-defined]
```

This patch aims to silence this error by ignoring the corresponding files, in order to satisfy CI. Let us know if you think this should be made happen differently.

With kind regards,
Andreas.

----

The root cause are those errors, where both https://github.com/python/mypy/issues/10301 and https://github.com/python/mypy/issues/5385#issuecomment-407281656 have interesting information, with a specific suggestion by @JelleZijlstra:

> I think you shouldn't make the protocol function `async def`, but just `def`. Conceptually, an async generator is a callable that returns an `AsyncIterator` (or more precisely, an `AsyncGenerator`). But an `async def` function without a `yield` returns an `Awaitable` of whatever its declared return type is, so that's how mypy interprets your protocol.

```
$ mypy --strict src/prompt_toolkit --platform win32
src/prompt_toolkit/completion/base.py:284: error: "_SupportsAcloseT" has no attribute "__aiter__" (not async iterable)  [attr-defined]
src/prompt_toolkit/completion/base.py:367: error: "_SupportsAcloseT" has no attribute "__aiter__" (not async iterable)  [attr-defined]
src/prompt_toolkit/completion/base.py:395: error: "_SupportsAcloseT" has no attribute "__aiter__" (not async iterable)  [attr-defined]
src/prompt_toolkit/buffer.py:1763: error: "_SupportsAcloseT" has no attribute "__aiter__" (not async iterable)  [attr-defined]
```

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -16,3 +16,9 @@ warn_redundant_casts = True
 warn_return_any = True
 warn_unused_configs = True
 warn_unused_ignores = True
+
+[mypy-prompt_toolkit.buffer]
+ignore_errors = True
+
+[mypy-prompt_toolkit.completion.base]
+ignore_errors = True


jonathanslenders:
Hi @amotl! Thanks for pointing out the issue.

I just had a look. No need to ignore anything.
This should fix it instead: https://github.com/prompt-toolkit/python-prompt-toolkit/pull/1711

amotl:
Lovely. Thanks for fixing it in a correct way.

PR #1721, opened 2023-02-01
Drop Python 3.6.

jonathanslenders:
Use `asyncio.get_running_loop()` and `asyncio.run()`. Stop using `asyncio.get_event_loop()` and `asyncio.set_event_loop()`.


jonathanslenders:
If anyone here has a Windows environment available, it would be much appreciated if this could be tested.
Windows 10 or 11 would be fine.

jonathanslenders:
Going to merge this, I tested on Windows.

There are some mypy issues: 
> AssertionError: Internal error: unresolved placeholder type None
That's probably a mypy bug which should be fixed in the next release: https://github.com/python/mypy/issues/14631

PR #1722, opened 2023-02-09
Partial suggestion filling for path parts

LoicGrobol:
This makes the ctrl+right partial suggestion filling take into account path parts, as it does in fish. To be precise, if the user has typed `python /home/l` and has been shown the suggestion `python /home/lgrobol/test.py`, hitting ctrl+right will only partial fill up to `python /home/lgrobol/` instead of  `python /home/lgrobol/test.py` (current behaviour).


jonathanslenders:
Thanks!

Issue #1688, opened 2022-11-15
[memory leak]PromptSession wouldn't free


vicalloy:
I use `PromptSession` with `prompt_toolkit.contrib.telnet`, and find it have serious  memory leak. `PromptSession` will never free.

1. Add `self.m = list(range(1000*1000))` to `prompt_toolkit.shortcuts.prompt.PromptSession.__init__`
2. Run the follow code. The memory increase fast and never free.

I try to find out the reason. `from prompt_toolkit.filters.base import _and_cache, _or_cache, _invert_cache` will hold `Filter`, and the `Filter` hold `PromptSession`. `prompt_toolkit.cache.memoized` also hold `PromptSession`.

I can free `_and_cache, _or_cache, _invert_cache`, but don't have any way to free the cache hold by `memoized`. I add a var `_memoized_cache_set` to access cache created by `memoized`. After clear `_and_cache/_or_cache/_invert_cache/memoized`, `PromptSession` will free. If called `await session.prompt_async()`, `PromptSession` wouldn't free again(only one instance can't free).

```python
import asyncio
import time
import gc

from prompt_toolkit import PromptSession


async def create_session():
    try:
        session = PromptSession()
        # await session.prompt_async()
    except KeyboardInterrupt:
        print("pass")
        del session
        gc.collect()


async def m():
    for _ in range(3):
        await create_session()
        time.sleep(3)

asyncio.run(m())
```

jonathanslenders:
That's a good find. This isn't great. We can't do without the memoization, or (CPU) performance will be worse.
I don't know what's best, either use some weak references, or have an `Filter._and_cache: _AndCache` attribute instead of a global. The latter could work.

jonathanslenders:
There should be something else.
I did an attempt to remove the global state in the filters here: https://github.com/prompt-toolkit/python-prompt-toolkit/pull/1690

But, even removing that cache completely + removing any cache from memoization elsewhere is not sufficient. I don't have much time right now, but I guess there's maybe another place that refers the session instance from a global state.

FYI, I'm using this to count the number of instances:
```python
import gc
obj = gc.get_objects()
print('len', len([o for o in obj if isinstance(o, PromptSession)]))
```

vicalloy:
Thanks for you reply.
I add a var `_memoized_cache_set` to access `@memoized`. After clean all cache, `PromptSession` can be free.
[...]

PR #1746, opened 2023-04-19
change filters's order to fix memory leak

vicalloy:
Reference: https://github.com/prompt-toolkit/python-prompt-toolkit/issues/1688

I updated to the latest `python-prompt-toolkit` and found that it still has memory leaks.
If a filter cached by a global filter function, such as `def is_done()`, the filter cannot be freed.  Filter operations should be very careful to avoid using global filters as the first filter.

Unresolved bug:
If I called `await session.prompt_async()`  one `PromptSession` is not freed.

```python
import sys
import asyncio
import gc
from prompt_toolkit import PromptSession

async def create_session():
    try:
        session = PromptSession()
        # await session.prompt_async()
    except KeyboardInterrupt:
        print("pass")
    del session
    gc.collect()

async def m():
    for _ in range(5):
        await create_session()
    gc.collect()
    obj = gc.get_objects()
    print('len', len([o for o in obj if isinstance(o, PromptSession)]))

asyncio.run(m())
```

### DIFF
--- a/src/prompt_toolkit/layout/menus.py
+++ b/src/prompt_toolkit/layout/menus.py
@@ -289,7 +289,7 @@ class CompletionsMenu(ConditionalContainer):
             ),
             # Show when there are completions but not at the point we are
             # returning the input.
-            filter=has_completions & ~is_done & extra_filter,
+            filter=extra_filter & has_completions & ~is_done,
         )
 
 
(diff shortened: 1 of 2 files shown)


jonathanslenders:
Thanks! The fix makes sense to me. I'm merging it, because it does fix the issue, and it's quite trivial. But long term, I have to think of a better solution.

So, the problem is that all filters have a cache within them, and if we do `global_filter & filter_that_references_our_prompt_session` then the global filter will hold a reference to our filter in the cache, which will hold a reference to the prompt session.

The cache is really important for performance, so it's quite tricky. I'm not sure what alternative we have.

Issue #1715, opened 2023-01-29
input_dialog  validation


sandrospadaro:
Validation does not work with input_dialog.

## how to reproduce the issue

```python
from prompt_toolkit.validation import Validator
from prompt_toolkit.shortcuts import input_dialog
from prompt_toolkit.validation import Validator


def is_number(text):
    return text.isdigit()

validator = Validator.from_callable(
    is_number,
    error_message='This input contains non-numeric characters')

age = input_dialog(title="Age", text="Your age:",
                   validator=validator).run()

print(f"Your age: {age}")
```

Fill the input dialog with some string (eg: abc) and press tab key to move focus on OK button. Then press enter key.

### Output

```bash
Your age: abc
```

### Output expected

Error message on focus lost

### About the environment

```bash
$ python -V
Python 3.11.1
```

```bash
$ pip freeze
prompt-toolkit==3.0.36
wcwidth==0.2.6
```

x24git:
Hi there, I created a PR that fixes the validation issue for input dialogs #1747. 
Until the PR is merged you can make your own `input_dialog` class that fixes the issue while using the base prompt-toolkit package

```python

class CustomValidationToolbar:
    def __init__(self, show_position: bool = False, buffer: Buffer = None) -> None:
        def get_formatted_text() -> StyleAndTextTuples:
            # If buffer not specified, use the currently focused buffer
            if buffer is None:
                buff = get_app().current_buffer
            else:
                buff = buffer
            if buff.validation_error:
                if show_position:
                    row, column = buff.document.translate_index_to_position(
                        buff.validation_error.cursor_position
                    )

                    text = "{} (line={} column={})".format(
                        buff.validation_error.message,
                        row + 1,
                        column + 1,
                    )
                else:
                    text = buff.validation_error.message

                return [("class:validation-toolbar", text)]
            else:
                return []

        self.control = FormattedTextControl(get_formatted_text)

        self.container = ConditionalContainer(
            content=Window(self.control, height=1), filter=True
        )

    def __pt_container__(self) -> Container:
        return self.container

def input_dialog(
    title: AnyFormattedText = "",
    text: AnyFormattedText = "",
    ok_text: str = "OK",
    cancel_text: str = "Cancel",
    completer: Completer | None = None,
    validator: Validator | None = None,
    validate_while_typing: FilterOrBool = False,
    password: FilterOrBool = False,
    style: BaseStyle | None = None,
    default: str = "",
[...]

PR #1747, opened 2023-04-20
Update Input Dialog to use Validator

x24git:
Currently, the Input Dialog accepts a Validator as a parameter but does not actually display anything via the UI. This PR fixes the following:

- Add option for `validate_while_typing` to `input_dialog` and passes along to TextField.
- Add a new `has_any_validation_error` filter that returns if any buffer in the application layout has a validation error.
- Modify `ValidationToolbar` to accept a specific buffer object rather than just the currently focused buffer.
- Modify the `input_dialog` object to validate when OK button is activated and prevent closing while validation errors exist.


x24git:
Should Fix Issue #1715

src/prompt_toolkit/filters/app.py
```diff
@@ -170,6 +171,33 @@ def has_validation_error() -> bool:
...
+    "Any buffer in the layout has validation error."
+    from prompt_toolkit.layout.containers import Window
+    from prompt_toolkit.layout.controls import (
+        BufferControl,
+        SearchBufferControl,
+        UIControl,
+    )
+
+    # Extract buffer validation errors for buffer UIControl children classes
+    def get_buffer_from_content(content: UIControl) -> bool:
```
jonathanslenders, line 185: Could you rename this? This doesn't return the buffer. For instance `buffer_has_validation_error()`

PR #1760, opened 2023-06-27
in_thread keyword is passed through 'prompt' shortcut

erwinkinn:
Hello!
First, thank you you all for such a powerful toolkit making absolutely easy to use interactive prompts. 
I develop an async code debugging tool based on clic_repl. I call click_repl session rigth from inside a coroutine.
click_repl in turn uses the toolkit to manage a prompt. It calls your `prompt_toolkit.shortcut.prompt.prompt` passing all the arguments as a dictionary. 
My issue is that I call it from inside a coroutine and I must pass `in_thread = True` because my loop is already  running. 
And this keyword is not passed further for some reason. 
This PR fixes it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,5 +45,8 @@ docs/_build
 # pycharm metadata
 .idea
 
+# vscode metadata
+.vscode
+
 # virtualenvs
 .venv*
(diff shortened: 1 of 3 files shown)


erwinkinn:
Hi, there! This repo has a `mypy` issue. I didn't do any castings but my PR (as well as the previous one) has some failed checks, hence I've added a fix.

jonathanslenders:
Thank you!

PR #1780, opened 2023-09-04
feat(keymap): add indent/dedent in insert mode

liljaylj:
`ctrl-t` to indent in insert mode
`ctrl-d` to dedent in insert mode

also, this commit place cursor after indent/dedent in the same place in the text instead of  placing it at the beginning of the line.


jonathanslenders:
Thanks!

jonathanslenders:
@liljaylj : I just noticed that there is still a bug. If there is no indentation left before the current line, pressing `c-d` will start moving the cursor to the left.

PR #1811, opened 2023-11-14
Fix IPython inputhook (workaround).

jonathanslenders:
Workaround to make IPython inputhooks work again, until https://github.com/ipython/ipython/pull/14241 is merged.

### DIFF
--- a/src/prompt_toolkit/application/application.py
+++ b/src/prompt_toolkit/application/application.py
@@ -956,10 +956,17 @@ class Application(Generic[_AppResult]):
             set_exception_handler=set_exception_handler,
             handle_sigint=handle_sigint,
         )
-        if inputhook is None:
-            # No loop installed. Run like usual.
-            return asyncio.run(coro)
-        else:
+
+        def _called_from_ipython() -> bool:
+            try:
+                return (
+                    "IPython/terminal/interactiveshell.py"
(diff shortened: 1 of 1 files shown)


tornaria:
This now causes a deprecation warning when running sagemath on python 3.12 with prompt_toolkit 3.0.41:
```
$ sage
┌────────────────────────────────────────────────────────────────────┐
│ SageMath version 10.1, Release Date: 2023-08-20                    │
│ Using Python 3.12.0. Type "help()" for help.                       │
└────────────────────────────────────────────────────────────────────┘
/usr/lib/python3.12/site-packages/prompt_toolkit/application/application.py:988: DeprecationWarning: There is no current event loop
  loop = asyncio.get_event_loop()
sage: 
```
But we are shipping ipython 8.18 (void linux) so I think this workaround is no longer required.

Is there any suggested way to move forward on this so that `asyncio.get_event_loop()` is not called when running python 3.12 and ipython 8.18 ? Are we expected to patch out this PR in our prompt_toolkit. Do you intend to do that in the future?

jonathanslenders:
@tornaria: Would you mind creating a pull request which adds the IPython version test? Maybe in `_called_from_ipython()` return `False` for newer IPython versions. For those we don't need the workaround any longer.

PR #1819, opened 2023-11-24
Bounded file history

robamu:
This limits the number of history entries which are loaded on a complete reload. It probably would also be possible to truncate the in-memory history, but I have not done that for now..


robamu:
Not exactly sure what the MyPy error is here. Looks like https://github.com/python/mypy/issues/12246 ..

PR #1823, opened 2023-12-07
Use `get_running_loop` instead of `get_event_loop`

auxsvr:
The latter is deprecated and may not always return the running event loop.


auxsvr:
`asyncio.get_event_loop` is already deprecated and may not return the running loop, cf. https://github.com/python/cpython/issues/96377#issuecomment-1260006538. According to the discussion there, it should be replaced with `asyncio.get_running_loop` or the corresponding policy method. If the `loop`  on line 988 is the running loop, this should not break old versions, if I understand the purpose of this code correctly. Of course, as I am not familiar with the codebase, I could be wrong.

auxsvr:
If I am wrong, please explain. At the moment, using `asyncio.get_running_loop` looks like the easiest option to me, preserving backward compatibility.

PR #1849, opened 2024-02-11
Expand type definition for filename

alexrudd2:
`FileHistory.filename` is currently hinted as `str`.

It is used by `open()` and `os.path.exists()`.  `os.path.exists()` supports `os.PathLike` [since 3.6](https://docs.python.org/3/library/os.path.html#os.path.exists)﻿

Therefore, the type hint should be expanded to `Union[str, os.PathLike]`.


jonathanslenders:
Thanks!

PR #1865, opened 2024-04-09
Fix compatibility when SIGINT handler is installed by non-Python

quark-zju:
When a native (Rust, C) program installs SIGINT handler and calls into Python, `signal.getsignal` can return `None`, which is not a valid value for the 2nd parameter of `signal.signal`. Fix it by checking `None` explicitly.


jonathanslenders:
Thanks!

PR #1880, opened 2024-06-04
Make AppSession compatible with pytest's capsys fixture + docs.

jonathanslenders:
Successor of https://github.com/prompt-toolkit/python-prompt-toolkit/pull/1878


tchalupnik:
Good one, thanks Jonatan! Works perfectly fine! :)

Issue #1648, opened 2022-07-02
Feature request: Word wrapping in buffers


VorpalBlade:
Hi,

I need to display a longer piece of text in my program. According to the docstrings FormattedTextControl/Label, is only suitable for shorter strings, so I went with Buffers/BufferControls, specifically the approach in #711 (for formatted buffer).

However neither that, nor the basic underlying buffer seems to support word wrapping (wrapping on space). Instead they wrap in the middle of words.

Either I'm missing something (in which case please inform me), or this is a feature request to support basic word wrapping.

I'm not asking for hyphenation, but it would be a nice bonus if this was implemented as a wrapping-policy. Then the user could provide a fancy version specific to their natural language of choice.

PR #1887, opened 2024-06-13
Word wrapping

Cimbali:
I wanted to use word wrapping and saw #1648  so here’s a preliminary PR, to discuss what needs changing/rewording/testing. In short, I’ve added a second callback similar to `GetLinePrefix` that defines wrapping points. They return a tuple of:
1. wrap position (self-explanatory)
2. number of chars to remove. Typically to remove the whitespace that was wrapped.\* You can also return -1 to truncate the line.
3. continuation text: any formatted text to insert before the line break to indicate continuation. Typically a hyphen if breaking a word in 2, or a `⮠` in a code block to indicate a non-semantic break, etc.

A default wrapping function is provided to do word wrapping (and it can even be re-used and tuned for different purposes). It can be activated with `word_wrap=True` on windows. For actual hyphenation a user would have to provide their own function.

\* E.g. if you want to wrap the text `example wrapped text` on 7 columns, you either break before the first space and remove it, or before and then your next line starts with a space (which looks weird). On 15 columns you’d have a similar problem but either break as `example ` / `wrapped ` / `text` without character removal, or `example wrapped` / `text`, so allowing character removal is good for efficiency of justified text.

### DIFF
--- a/examples/prompts/auto-completion/fuzzy-custom-completer.py
+++ b/examples/prompts/auto-completion/fuzzy-custom-completer.py
@@ -36,21 +36,28 @@ class ColorCompleter(Completer):
 def main():
     # Simple completion menu.
     print("(The completion menu displays colors.)")
-    prompt("Type a color: ", completer=FuzzyCompleter(ColorCompleter()))
-
-    # Multi-column menu.
-    prompt(
+    r = prompt(
         "Type a color: ",
         completer=FuzzyCompleter(ColorCompleter()),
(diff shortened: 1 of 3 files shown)


Cimbali:
I’ve added an example that showcases the feature, here’s what it looks like with:
1) no wrapping
2) default word wrapping
3) customised word wrapping with line continuation markers
4) truncation (with a "..." marker)

image

PR #1889, opened 2024-06-16
Update type definition for `words` argument

mxr:
The type definition permits `list[str] | Callable[[], list[str]]` however a tuple is valid too. Updating the type definition to use `Sequence[str] | Callable[[], Sequence[str]]` to allow using a tuple without errors


jonathanslenders:
Thanks!

PR #1900, opened 2024-07-29
Support various cursor shapes for more input modes

Freed-Wu:
Related: https://github.com/prompt-toolkit/python-prompt-toolkit/issues/192


jonathanslenders:
Thanks!

jonathanslenders:
Two comments here:
- There was a bug in this PR that I fixed here: https://github.com/prompt-toolkit/python-prompt-toolkit/pull/1920
- Displaying a "beam" as a cursor for `INSERT_MULTIPLE` does not work currently, because we can't display multiple cursors. So it renders instead by coloring the background of these cells.

PR #1948, opened 2024-12-19
Minimal changes to allow multiline suggestions.

Carreau:
I'm trying to be the least invasive;
But the idea is to allow access to previous lines when rendering a line in order to shift them down; and display multiline sugestions.

With this I can add minimal modifications into IPython that allows multiline suggestions (typically would be created by an LLM – while not a big fan, there is a lot of requests for it.

I'm going to assume there is no willingness tochange the current API for transformers to get at least a ranges of lines instead of lines, it would be easier to handle.

I do not include any update to the transformers we use in IPython to display those completions; but once battled tested a bit more, I'm happy to contribute it as well.

---- 

Here is what it could looks like in IPython for example. 

https://github.com/user-attachments/assets/39304c01-7461-45f8-8d96-607a8c558194

### DIFF
--- a/src/prompt_toolkit/layout/controls.py
+++ b/src/prompt_toolkit/layout/controls.py
@@ -667,7 +667,11 @@ class BufferControl(UIControl):
 
         merged_processor = merge_processors(input_processors)
 
-        def transform(lineno: int, fragments: StyleAndTextTuples) -> _ProcessedLine:
+        def transform(
+            lineno: int,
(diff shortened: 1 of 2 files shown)


jonathanslenders:
Thank you!

Carreau:
Thanks, and happy new year. 

I'll contribute some examples that show how one can achieve showing multiple lines later this month.

PR #1953, opened 2025-01-07
Basic multiline suggestion example

Carreau:
Unlike inline completion; this is a bit more tricky to set up, in particular it is hard to choose default that will suit everyone – whether to shift existing line, how to accept/reject, show elision...

So we just for now add an example on how this can be used.

I will most likely make use of it in IPython in the next few weeks/month, and  can report back on the usability.

--- 

This follows #1948, which enable multiline suggestion. Here I show how to use it with a custom preprocessor that will try to fill in multiline text by pushing the existing text down when rendering.

https://github.com/user-attachments/assets/8924087f-a45c-4e67-ab20-270873bf7606

This is of course just an example as it does not handle accepting the suggestion; or marking that text has been pushed past the end of the buffer and is not shown.

All this rely on the `return Transformation(ti.get_line(shift))`, as `get_line` is only available since #1948

### DIFF
--- a/examples/prompts/multiline-autosuggest.py
+++ b/examples/prompts/multiline-autosuggest.py
@@ -1,0 +1,154 @@
+#!/usr/bin/env python
+"""
+A more complex example of a CLI that demonstrates fish-style auto suggestion
+across multiple lines.
+
+This can typically be used for LLM that may return multi-line responses.
+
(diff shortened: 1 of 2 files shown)


jonathanslenders:
Thank you! Really appreciated!

I hope to do a release in the coming days.

PR #1954, opened 2025-01-08
Try to run ruff and typo only on 3.13

Carreau:
There is no need to run N times.

This in particular means that we can still see from a glance the tests are passing on older python.

I also try to run typo regardless of wether ruff check and formatting passes, which hopefully avoid one iteration of fix, push and CI failing.


Carreau:
I'm not the best at GitHubb action syntax though.

jonathanslenders:
Thanks!

jonathanslenders:
I think something went wrong here. GitHub actions is no longer running since I merged this.

jonathanslenders:
Fixed here: https://github.com/prompt-toolkit/python-prompt-toolkit/pull/1965

PR #1962, opened 2025-01-17
Hide password during typing with a new prompt input

### LINKED ISSUE #1783: Hide password instead of replacing it with a character
Hey there, `questionary` user here.

I opened this issue to hide the password that an user types when using `questionary.password()` (a wrapper around `PromptSession(is_password=True)`: https://github.com/tmbo/questionary/issues/321

Would be nice to specify the character to use to mask the password; right now this is `*`, and from a quick look it doesn't seem possible to change this behavior from an external user point of view? But correct me if I'm wrong obviously.

Line of code affected: https://github.com/prompt-toolkit/python-prompt-toolkit/blob/490cf900188df357611cdc9358256be0e5ce2e16/src/prompt_toolkit/layout/processors.py#L311

Would be nice to select either the character to use, e.g., empty string `""`, or just have a switch to mask the password at all, leaving the default `*` in case the password is not masked.

Crissal1995:
This PR resolves #1783.
The `PasswordProcessor` by default (current implementation, no argument given) outputs a `*` for each letter typed by the user in the prompt.
However, the desired behavior would be to make the typed password invisible, e.g. like `getpass` does.
Two tests were added, and the example was also modified and tested locally to make sure that the change works as intended.

### DIFF
--- a/src/prompt_toolkit/shortcuts/prompt.py
+++ b/src/prompt_toolkit/shortcuts/prompt.py
@@ -249,6 +249,7 @@ class PromptSession(Generic[_T]):
         When True (the default), automatically wrap long lines instead of
         scrolling horizontally.
     :param is_password: Show asterisks instead of the actual typed characters.
+    :param hide_password: Hide the password, rather than showing asterisks.
     :param editing_mode: ``EditingMode.VI`` or ``EditingMode.EMACS``.
     :param vi_mode: `bool`, if True, Identical to ``editing_mode=EditingMode.VI``.
     :param complete_while_typing: `bool` or
@@ -335,10 +336,10 @@ class PromptSession(Generic[_T]):
         "lexer",
         "completer",
         "complete_in_thread",
-        "is_password",
         "editing_mode",
         "key_bindings",
         "is_password",
+        "hide_password",
         "bottom_toolbar",
         "style",
         "style_transformation",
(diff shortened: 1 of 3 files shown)


src/prompt_toolkit/shortcuts/prompt.py
```diff
@@ -249,6 +249,7 @@ class PromptSession(Generic[_T]):
         When True (the default), automatically wrap long lines instead of
         scrolling horizontally.
     :param is_password: Show asterisks instead of the actual typed characters.
+    :param hide_password: Hide the password, rather than showing asterisks.
```
jonathanslenders, line 252: Would it make sense to accept `password_character` here instead of `hide_password`? It will be more flexible.

Crissal1995: Undoubtedly, but from an user perspective `hide_password` is more clear.

Besides, is there any valuable use-case where the password is displayed as something else, e.g. `xxxxx` or `11111`?

Crissal1995:
Apparently running the tests with `pytest` leads to different results than the command used by the CI pipeline, `coverage run -m pytest`; I am not sure why this is the case.

src/prompt_toolkit/shortcuts/prompt.py
jonathanslenders: I can imagine that some might want to render it as Unicode bullets, because that's more beautiful.

Like `password_character="•"`

I mainly want to avoid having too many redundant arguments here. We have to repeat them in multiple places and the signature gets pretty long. If at some point, some users want to configure this, we'll have both `hide_password` and `password_character`.

Honestly, I'm not sure myself. If there was some prior art in other libraries, that could convince me to go one way or another.

Crissal1995: An alternative would be to use both flags, forcing `password_character` to be an empty string if `hide_password` is `True`.
In this way users could either use the flag `hide_password` if they don't care about how the password is displayed on screen, or the flag `password_character` if they want to customize the appearance, as youy suggested.

PR #1963, opened 2025-01-17
Fix YAML syntax error

Crissal1995:
This PR fixes https://github.com/prompt-toolkit/python-prompt-toolkit/actions/runs/12838621403


jonathanslenders:
Oh thanks! I missed this PR and did the same fix here: https://github.com/prompt-toolkit/python-prompt-toolkit/pull/1965

PR #1965, opened 2025-01-20
Fix Github Actions.

jonathanslenders:


Carreau:
Hum, sorry about that. I was trying to get typos to run even if previous one was failing.

PR #1982, opened 2025-04-10
Fix edge case with leading line ending in 'split_lines'.

jonathanslenders:


joouha:
Hi Jonathan,
I was just wondering what was the rational behind this change? 
From my perspective, it results in the generation of unnecessary extra style-and-text tuples around new-lines with no content.

PR #1986, opened 2025-04-16
Re-add project URLs after migration to pyproject.toml

Jamie-:
Following spec:
https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#urls

The previous URL was removed in https://github.com/prompt-toolkit/python-prompt-toolkit/commit/88a011ccde71fe1df6b8e873801d1276859e9e70 which has removed all URLs from PyPI.

The new format allows listing of different types of URL so have added the most common ones.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -25,6 +25,11 @@ dependencies = [
     "wcwidth",
 ]
 
+[project.urls]
+homepage = "https://github.com/prompt-toolkit/python-prompt-toolkit"
+source = "https://github.com/prompt-toolkit/python-prompt-toolkit"
+documentation = "https://python-prompt-toolkit.readthedocs.io/en/stable/"
+
 [tool.ruff]
 target-version = "py37"
 lint.select = [


pyproject.toml
```diff
@@ -25,6 +25,11 @@ dependencies = [
     "wcwidth",
 ]
 
+[project.urls]
+homepage = "https://github.com/prompt-toolkit/python-prompt-toolkit"
+source = "https://github.com/prompt-toolkit/python-prompt-toolkit"
+documentation = "https://python-prompt-toolkit.readthedocs.io/en/stable/"
```
cdce8p, line 31:
```suggestion
Homepage = "https://github.com/prompt-toolkit/python-prompt-toolkit"
Documentation = "https://python-prompt-toolkit.readthedocs.io/en/stable/"
```
A minor suggestion. Not sure there is a need to duplicate urls.

Jamie-: They show under different headings in PyPI hence I included both so you get both links. Happy to adjust though if you would rather simplify?

cdce8p: True. Though I wonder how useful it would be to have to different headings link to the same page.

I'd also suggest to capitalize the keys.

Jamie-: Have updated accordingly

jonathanslenders:
Thanks!

PR #1987, opened 2025-04-16
Fix images in README.rst

Jamie-:
Github images can't be linked to directly any more. The recommended method is to use repository relative URLs.

You can see this working [here](https://github.com/Jamie-/fork-python-prompt-toolkit/blob/fix-readme-images/README.rst)


jonathanslenders:
Thanks!

PR #1998, opened 2025-06-11
use correct width of non-english character in `Button`

python-and-novella:
The test code is:
```python3
from prompt_toolkit.shortcuts import message_dialog

message_dialog(
    title='简单对话框',
    text='回车或者点击确认',
    ok_text='确认').run()
```
It should show `Button`'s `right_symbol`:
image


jonathanslenders:
Thank you @python-and-novella

PR #2001, opened 2025-07-19
Implement flushing in Windows VT100 input

NewDefectus:
The flushing logic for VT100 parsing was already there, it just wasn't called by the Windows input reader for some reason.

For transparency - I copied the `flush_keys` method verbatim from `vt100.py`.


jonathanslenders:
Thank you!

PR #2002, opened 2025-08-04
Update docs theme to Nefertiti 0.8.6

danirus:
* Use variable font Ubuntu Sans Mono and adapt font size.
* Table Of Content (right side panel) for the reference API page will benefit from last fix included in 0.8.6.


jonathanslenders:
Thanks!

PR #2063, opened 2026-04-07
FuzzyCompleter: compile re pattern on init

bretello:
- remove duplicated assignement
- compile re on init to avoid compiling the re on every completion request


bretello:
CI failure is unrelated, see https://github.com/prompt-toolkit/python-prompt-toolkit/pull/2064

PR #2066, opened 2026-04-15
Optionaly recognize modifyOtherKey (ctrl-, ctrl-shift-)

Carreau:
As far as I can tell, many terminal support this,
and it allows (when enabled) to distinguish some modifier that are otherwise impossible to detect.

In particular there is often a request in IPython to distinguish Ctrl-Enter.

This _should_ be 100% backward compatible.

Ctrl-Shift- is _not_ added as many terminal don't actually support it.

This is gated by a context manager `modify_other_keys` that is widely supported.

Some discussion I found about why / or why not :

`https://github.com/ghostty-org/ghostty/issues/242`

related to https://github.com/ipython/ipython/issues/13879

--- 

I tried state 2 but this would be much more invasive to detect all the keys, any anyway if we go further I think support for https://sw.kovidgoyal.net/kitty/keyboard-protocol/ directly will be good as there is extensive documentation on how to detect it and have progressive enhancement.

Leaving as draft for now

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,20 @@
 CHANGELOG
 =========
 
+Unreleased
+----------
+
+New features:
+- Distinguish Ctrl-Enter and Ctrl-Shift-Enter from plain Enter on terminals
+  that support xterm's `modifyOtherKeys` protocol. The protocol is enabled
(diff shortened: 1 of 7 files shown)


Carreau:
Havind this and kitty protocol at the same time is a bit annoying, maybe I'll send something that just adds kitty protocol

PR #2067, opened 2026-04-16
Support the Kitty keyboard protocol (flag 1, disambiguate)

Carreau:
Push the Kitty keyboard protocol on startup and pop it on exit so
supporting terminals deliver modified keys — Ctrl-Enter, Shift-Enter,
Ctrl-Shift-Enter, Alt-letter, and full modifier coverage on the
navigation block — as distinct CSI u sequences instead of being
collapsed into their unmodified equivalents. Terminals that don't
implement the protocol silently ignore the push, so there is no
regression for them.

Detection mirrors the existing CPR machinery: the renderer emits
`CSI ? u`, a new binding under key_binding/bindings/ consumes the
response and flips `Renderer.kitty_support` from UNKNOWN to SUPPORTED,
letting callers branch on capability if they want.

The bulk of the code lives in new files:
- input/kitty_keyboard.py — CSI u decoder, functional-key table,
  query-response parser
- output/kitty_keyboard.py — push/pop context manager with
  reference-counted depth, sequence constants
- key_binding/bindings/kitty_keyboard.py — response-consuming binding
- docs/pages/advanced_topics/kitty_keyboard_protocol.rst — maintainer
  notes on wire format, capability detection, and known sharp edges

Touch-points in existing code are small and mirror the CPR pattern:
one import + one regex + one dispatch branch in vt100_parser.py, a
push/query/pop trio in renderer.py, one binding registration in
key_binding/defaults.py, four new enum values in keys.py.

prompt_toolkit does not push xterm's modifyOtherKeys. Users whose
terminal or tmux has it enabled independently still get the existing
`CSI 27` Enter fallback in ansi_escape_sequences.py, which folds all
three modifier+Enter variants to plain Keys.ControlM so the form at
least submits rather than silently doing nothing.

New Keys: ControlEnter, ControlShiftEnter, ShiftEnter,
KittyKeyboardResponse. Bindings registered as `c-enter` / `s-enter` /
`c-s-enter` fire on Kitty-capable terminals; on non-Kitty terminals
they don't fire (plain Enter fires instead, as before).

### DIFF
--- a/docs/pages/advanced_topics/kitty_keyboard_protocol.rst
+++ b/docs/pages/advanced_topics/kitty_keyboard_protocol.rst
@@ -58,11 +58,15 @@ Input
 decoder. Covered:
 
 - Functional keys from the Kitty spec: :kbd:`enter`, :kbd:`tab`,
-  :kbd:`escape`, :kbd:`backspace`, arrows, navigation block
-  (:kbd:`home` / :kbd:`end` / :kbd:`pageup` / :kbd:`pagedown` /
-  :kbd:`insert` / :kbd:`delete`), :kbd:`f1`–:kbd:`f12`. Mapped to the
+  :kbd:`escape`, :kbd:`backspace`, :kbd:`f1`–:kbd:`f12`. Mapped to the
   nearest existing ``Keys`` value with Shift / Ctrl / Ctrl-Shift
-  promotion where an enum exists.
+  promotion where an enum exists. Arrow keys and the navigation block
+  (:kbd:`home` / :kbd:`end` / :kbd:`pageup` / :kbd:`pagedown` /
+  :kbd:`insert` / :kbd:`delete`) are **not** handled here — under
+  flag 1 the Kitty spec keeps them in their legacy
+  ``CSI <n> ~`` / ``CSI <letter>`` / ``SS3 <letter>`` encoding even
+  when modified, so they continue to travel through
+  ``ANSI_SEQUENCES`` (which already has the full modifier matrix).
 - Printable Unicode keys with Ctrl (mapped to ``Keys.ControlX``) and
   Ctrl+Shift digits (mapped to ``Keys.ControlShift1`` …).
 - Alt as a meta prefix: emitted as ``(Keys.Escape, base_key)`` to match
@@ -120,14 +124,15 @@ New ``Keys`` values
 - ``Keys.ControlEscape``, ``Keys.ControlShiftEscape`` — Kitty-only
   modifier+Escape distinctions, alongside the pre-existing
   ``Keys.ShiftEscape``. Same non-Kitty fallback behavior.
+- ``Keys.ControlBackspace``, ``Keys.ShiftBackspace``,
+  ``Keys.ControlShiftBackspace`` — Kitty-only modifier+Backspace
+  distinctions. Unlike modified Enter, there is no safe legacy
+  fallback: on most non-Kitty terminals Ctrl-Backspace is
+  indistinguishable from plain Backspace or from Ctrl-H, so we do
+  not fold these down — a binding on one of them will simply not
+  fire on non-Kitty terminals.
 - ``Keys.KittyKeyboardResponse`` — internal sentinel for the query
   response parser-to-binding dispatch.
-
-Backspace (:kbd:`backspace`) is decoded from the Kitty functional-key
-code 127, but prompt_toolkit has no distinct ``Keys`` values for
-:kbd:`c-backspace` / :kbd:`s-backspace`, so modified Backspace silently
-folds back to plain Backspace (``Keys.ControlH``) — same behavior as on
-legacy terminals.
 
 
 What could be done in the future
--- a/docs/pages/advanced_topics/kitty_keyboard_protocol.rst
+++ b/docs/pages/advanced_topics/kitty_keyboard_protocol.rst
@@ -1,0 +1,244 @@
+.. _kitty_keyboard_protocol:
+
+
+Kitty keyboard protocol
+=======================
+
+Maintainer-facing notes on prompt_toolkit's support for the `Kitty
+keyboard protocol <https://sw.kovidgoyal.net/kitty/keyboard-protocol/>`_.
+
+Despite its name, the Kitty protocol is supported by a wide range of
+terminal emulators across platforms and is not limited to the Kitty
+terminal itself.
+
+Only flag 1 ("disambiguate escape codes") is currently implemented.
+The spec also defines progressive-enhancement flags for reporting
+press/release/repeat events, alternate keys, all keys as escape codes,
+and associated text; none of those are implemented here.
+
+
+Why
+---
+
+Under legacy terminal encodings, many modifier+key combinations are
+ambiguous or impossible — :kbd:`c-enter` sends the same ``\r`` as plain
+:kbd:`enter`, :kbd:`s-enter` is indistinguishable from :kbd:`enter` on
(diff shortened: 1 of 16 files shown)


Carreau:
Alternate to #2066, that push it (much further) using kitty, draft for now, I need to check some things, and trim it down as it only support part of the kitty protocol; I need to make sure of a few things (whether it's set or push/pop; and how it behaves in some case) and re-read everything

joouha:
Hi Carreau, Jonathan,

I have also implemented support for CSI-u / kitty keyboard protocol in `prompt_toolkit` for euporie, but I took a slightly different approach to Claude's implementation here, which might be of interest. 

I generated new entries for [`prompt_toolkit.keys:Keys`](https://github.com/joouha/euporie/blob/d370c0f592f7039607d38b87ca6319b32de37a04/packages/apptk/src/apptk/keys.py#L8) and [`prompt_toolkit.input.ansi_escape_sequences:ANSI_SEQUENCES`](https://github.com/joouha/euporie/blob/d370c0f592f7039607d38b87ca6319b32de37a04/packages/apptk/src/apptk/input/ansi_escape_sequences.py) using [this script](https://github.com/joouha/euporie/blob/d370c0f592f7039607d38b87ca6319b32de37a04/scripts/generate_more_keys.py), adding kitty and CSI-u keys and sequences. This way, the vt100 input parser can be used as is, without the need to add special handling for kitty style input sequences (they are just treated like any other key-press). Having a greatly expanded list of recognised keys does not appear to have a noticeable performance impact.

I also added [`enable/disable_extened_keys`](https://github.com/joouha/euporie/blob/d370c0f592f7039607d38b87ca6319b32de37a04/packages/apptk/src/apptk/output/vt100.py#L326-L338) methods to the `Output` class (only implemented for `Vt10Output`). Rather than keeping track of the kitty keyboard state, I simply push and pop from the stack before and after each render cycle (so more akin to how mouse support is implemented). 

The `apptk` package I have linked to above is a [`modshim`](https://github.com/joouha/modshim) overlay package for `prompt_toolkit`, which contains many extensions and new features. I'd be happy to split out the keyboard input changes (or anything else) into an alternative to this PR if that would be useful.

Carreau:
Ah, that is interesting! i will have a look!

Carreau:
I'd love to know if @jonathanslenders would be open to having this in core anyway; otherwise I'll if I can adopt a method similar to yours in IPython (with credit).

Carreau:
Just nudging @jonathanslenders to know if this is something I  should be working on, or trying to work around in IPython.

jonathanslenders:
Yes! I'm open to supporting the Kitty keyboard protocol.

I have not had the time yet to review this due to lack of time. This is quite a big PR.

(Is this an AI contribution btw? Just asking.)

Carreau:
Claude helped yes; 

i wanted to make sure you were ok for support; if it's not ok to have AI help I can completely redo it by hand. But i wanted to know if you were ok with kitty in prompt toolkit before pushing on it.

Take your time to focus on other things; i'll mark as ready for review once i actually want a review :-)

jonathanslenders:
Claude is fine as long as the code is properly reviewed. Also, I'm definitely fine with the Kitty protocol, it was even on my wish list of things I'd love to have.

docs/pages/advanced_topics/kitty_keyboard_protocol.rst
```diff
@@ -0,0 +1,253 @@
...
+  ``kitty_support`` flips to ``SUPPORTED``. That only affects the
+  capability signal — the push itself applies immediately, so the
+  terminal's first keystroke is already in the new encoding.
+- **Functional-key codes are not universal.** The Kitty spec pins
+  Enter=13 (which coincides with ``\r``) but implementations disagree
+  on some rarer functional codes. Worth spot-checking new ones against
+  kitty, ghostty, wezterm, foot.
+- **Alt vs. Esc-prefix.** Kitty reports Alt as a modifier; the legacy
+  path reports it as ``(Esc, letter)``. The decoder emits
+  ``(Keys.Escape, base_key)`` for Alt-prefixed keys to match legacy
```
jonathanslenders, line 251: This is fine btw. It's similar to how things work on Windows.

src/prompt_toolkit/output/kitty_keyboard.py
```diff
@@ -0,0 +1,114 @@
...
+
+    A nested call must use the same `flags` as the outer one. Mixing
+    flags would leave the terminal in a state the inner caller didn't
+    ask for, and the inner pop would cancel the outer push; we raise
+    `ValueError` rather than silently corrupt the stack.
+    """
+    # Lazily install counter + flags state on the Output. We do it here
+    # rather than on Output.__init__ so this module stays fully
+    # self-contained and the rest of the codebase has no mention of it.
+    depth: int = getattr(output, "_kitty_keyboard_depth", 0)
```
jonathanslenders, line 92: In general, I avoid any usage of `getattr`/`AttributeError`/`setattr`/`hasattr`. This does not allow for type checking.

Better use polymorphism, and ensure that the `Output` interface has methods for anything we need. Don't assume attributes on an interface, only methods.

src/prompt_toolkit/renderer.py
```diff
@@ -362,6 +373,15 @@ def __init__(
         self._in_alternate_screen = False
         self._mouse_support_enabled = False
         self._bracketed_paste_enabled = False
+        # ExitStack holding the active kitty_keyboard_protocol context (or
+        # empty if the protocol isn't currently pushed). Using a stack
+        # instead of a manual __enter__/__exit__ pair means `close()` is
+        # a no-op when nothing is registered — no null-check needed — and
+        # an exception during `__enter__` is surfaced by the stack itself
+        # rather than leaving a half-entered CM on the renderer.
+        self._kitty_keyboard_stack: ExitStack = ExitStack()
```
jonathanslenders, line 382: I'm not a big fan of using `ExitStack` in a way where it's not being used as a context manager, because there's no guarantee that the stack will be closed.

However, the right approach (if we need `ExitStack`) would be to add a classmethod constructor to `Renderer` which behaves as a contextmanager, and as a result of that, we'd need to do the same for `Application`. This is not feasible without breaking APIs.

Maybe my thinking here is a result of my focus on structured concurrency over the past 5 years or so.

I need to take some time to review in more detail, but it might be better to not implement `kitty_keyboard_protocol` as a contextmanager, but as distinct methods on the `Output` implementation:

- `Output.enable_kitty_protocol()` and `Output.disable_kitty_protocol()`

The `Vt100Output` implementation can then have a field `_kitty_keyboard_depth` and keep track of that.

src/prompt_toolkit/renderer.py
```diff
@@ -609,6 +641,44 @@ def render(
...
+            # Query first, then push. The response to `CSI ? u` reports
+            # the currently-active flags, so emitting it before the push
+            # keeps the response reflective of the terminal's pre-push
+            # state (a future version that reads the response's flags to
+            # decide whether to push at all needs this ordering). Skip
+            # the query when we already know the answer — re-asking on
+            # every render burns a round-trip and has no value: the
+            # underlying terminal doesn't change capability under us.
+            if self.kitty_support is KittySupport.UNKNOWN:
+                self.output.write_raw(KITTY_QUERY)
```
jonathanslenders, line 658: Better add a method `Output.ask_for_kitty_support()` similar to `Output.ask_for_cpr()`. `Output` implementations that don't support Kitty don't need to implement that. (empty method in base class is fine).

This renderer class should not be concerned with specific Vt100 escape sequences like `KITTY_QUERY`.

PR #2070, opened 2026-05-10
Fix FuzzyWordCompleter meta_dict type hint to match WordCompleter

### LINKED ISSUE #2039: FuzzyWordCompleter type hint bug
https://github.com/prompt-toolkit/python-prompt-toolkit/blob/c7c629c38b22b5f22a0a5e80a1ae7e758ac595bb/src/prompt_toolkit/completion/fuzzy_completer.py#L191

Perhaps it should be `meta_dict: Mapping[str, AnyFormattedText] | None = None`

bibekmhj:
Fixes #2039.

`FuzzyWordCompleter` types `meta_dict` as `dict[str, str] | None` and then forwards it straight into `WordCompleter`, which declares the same parameter as `Mapping[str, AnyFormattedText] | None`. So the wrapper rejects values the wrapped class would happily accept — passing in something like `{"alpha": HTML("<b>first</b>")}` works at runtime but trips the type checker.

Widened it to match `WordCompleter`. `AnyFormattedText` is already imported in this file, just needed to add `Mapping` from typing. No runtime behavior change.

### DIFF
--- a/src/prompt_toolkit/completion/fuzzy_completer.py
+++ b/src/prompt_toolkit/completion/fuzzy_completer.py
@@ -1,8 +1,8 @@
 from __future__ import annotations
 
 import re
-from collections.abc import Callable, Iterable, Sequence
-from typing import NamedTuple, Mapping
+from collections.abc import Callable, Iterable, Mapping, Sequence
+from typing import NamedTuple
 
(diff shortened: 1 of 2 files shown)


jonathanslenders:
Thanks!

Issue #1991, opened 2025-05-08
examples/ssh /asyncssh-server.py a CTRL-C during the progress kills also the Server


gierig1977:
Starting to Discover async, AsyncSSH and as well that very nice prompt_toolkit

Demo (examples/ssh /asyncssh-server.py) seems perfect as "StartPoint" to discover.
Demo is running fine. 
But when you Press "CTRL-C" During the  progress bar demo it will also kill the Server.

latest Trackack:
```
Traceback (most recent call last):
  File "/opt/homebrew/Cellar/python@3.13/3.13.3/Frameworks/Python.framework/Versions/3.13/lib/python3.13/asyncio/runners.py", line 118, in run
    return self._loop.run_until_complete(task)
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^
  File "/opt/homebrew/Cellar/python@3.13/3.13.3/Frameworks/Python.framework/Versions/3.13/lib/python3.13/asyncio/base_events.py", line 719, in run_until_complete
    return future.result()
           ~~~~~~~~~~~~~^^
  File "/Users/gierig/Library/Mobile Documents/com~apple~CloudDocs/Documents/code/myBBS/test.py", line 117, in main
    await asyncio.Future()
asyncio.exceptions.CancelledError

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/opt/homebrew/Cellar/python@3.13/3.13.3/Frameworks/Python.framework/Versions/3.13/lib/python3.13/asyncio/runners.py", line 195, in run
    return runner.run(main)
           ~~~~~~~~~~^^^^^^
  File "/opt/homebrew/Cellar/python@3.13/3.13.3/Frameworks/Python.framework/Versions/3.13/lib/python3.13/asyncio/runners.py", line 123, in run
    raise KeyboardInterrupt()
KeyboardInterrupt

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/gierig/Library/Mobile Documents/com~apple~CloudDocs/Documents/code/myBBS/lib/python3.13/site-packages/prompt_toolkit/contrib/ssh/server.py", line 103, in _interact
    await self.interact(self)
  File "/Users/gierig/Library/Mobile Documents/com~apple~CloudDocs/Documents/code/myBBS/test.py", line 77, in interact
    await asyncio.sleep(0.1)
[...]

PR #2072, opened 2026-05-18
Handle asyncio.CancelledError in SSH session interaction

dnyanesh1011:
Fixes noisy traceback propagation during SSH session cancellation.

Currently `_interact()` catches `BaseException`, which also captures `asyncio.CancelledError`. In modern asyncio versions (especially Python 3.11+ / 3.13), `CancelledError` is expected control-flow during task shutdown and should not be treated as an application error.

This change:

* handles `asyncio.CancelledError` explicitly
* keeps normal application exceptions visible
* avoids traceback spam during SSH disconnect/session cancellation
* improves cancellation behavior when pressing `Ctrl-C` during the asyncssh example progress demo

Related issue: #1991

### DIFF
--- a/src/prompt_toolkit/contrib/ssh/server.py
+++ b/src/prompt_toolkit/contrib/ssh/server.py
@@ -34,13 +34,11 @@ class PromptToolkitSSHSession(asyncssh.SSHServerSession):  # type: ignore
         self.app_session: AppSession | None = None
 
         # PipInput object, for sending input in the CLI.
-        # (This is something that we can use in the prompt_toolkit event loop,
-        # but still write date in manually.)
+        # (This is something that we can use in the prompt_toolkit event loop, but still write data in manually.)
         self._input: PipeInput | None = None
         self._output: Vt100_Output | None = None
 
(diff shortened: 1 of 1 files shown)


src/prompt_toolkit/contrib/ssh/server.py
```diff
@@ -100,10 +100,17 @@ async def _interact(self) -> None:
...
             with create_app_session(input=self._input, output=self._output) as session:
                 self.app_session = session
+
                 try:
                     await self.interact(self)
-                except BaseException:
+
+                except asyncio.CancelledError:
+                    # Expected during disconnect/shutdown.
+                    pass
```
jonathanslenders, line 109: Can you propagate rather than suppress using `raise` instead of `pass`?

dnyanesh1011: Good catch, I updated the handler so asyncio.CancelledError is now re-raised instead of being swallowed.   Regular application exceptions are still handled separately, so cancellation behavior stays intact without mixing it into normal error handling

PR #2077, opened 2026-07-19
Ensure confirm() returns a boolean after Enter

### LINKED ISSUE #2046: prompt_toolkit.shortcuts.confirm doesn't always return bool
`prompt_toolkit.shortcuts.confirm` doesn't always return `bool`. When pressing just enter it returns a `str`.

For example:
```python
from prompt_toolkit.shortcuts import confirm

result = confirm()

print(f"Result|{result}|")
print(type(result))
```
Produces:
```
Confirm? (y/n)  ### I press `Enter` here ###
Result||
<class 'str'>
```
This is on Windows, using prompt_toolkit 3.0.52 if it matters.

zhangli091011:
## Summary

Fix `prompt_toolkit.shortcuts.confirm()` returning an empty string when Enter is pressed without selecting an answer.

Enter previously fell through to `PromptSession`'s normal accept binding, which returned the empty buffer contents. The confirmation session now handles Enter explicitly and continues waiting for `y` or `n`, so successful results remain `True` or `False`.

The CI environment's current mypy/typeshed versions also exposed an incompatible return annotation in the existing custom PDB trace dispatcher. Its internal return annotation is now `Any`, matching the range of values accepted by the debugger API without changing runtime behavior.

Fixes #2046.

## Tests

- Added regression coverage for Enter followed by `y`, `Y`, `n`, and `N`
- `.venv\Scripts\python.exe -m pytest tests/test_shortcuts.py -q -k confirm`
- `ruff check src/prompt_toolkit/shortcuts/prompt.py tests/test_shortcuts.py`
- `ruff format --check src/prompt_toolkit/shortcuts/prompt.py tests/test_shortcuts.py`
- Strict mypy check for the changed implementation
- `typos .`

## AI disclosure

OpenCode using `myself/gpt-5.6-sol` assisted with issue investigation, implementation, and test preparation. I reviewed the change and verification results.

### DIFF
--- a/src/prompt_toolkit/application/application.py
+++ b/src/prompt_toolkit/application/application.py
@@ -1055,8 +1055,6 @@ class Application(Generic[_AppResult]):
         import pdb
         from types import FrameType
 
-        TraceDispatch = Callable[[FrameType, str, Any], Any]
-
         @contextmanager
         def hide_app_from_eventloop_thread() -> Generator[None, None, None]:
             """Stop application if `__breakpointhook__` is called from within
@@ -1110,9 +1108,7 @@ class Application(Generic[_AppResult]):
                 done.set()
 
         class CustomPdb(pdb.Pdb):
-            def trace_dispatch(
-                self, frame: FrameType, event: str, arg: Any
-            ) -> TraceDispatch:
+            def trace_dispatch(self, frame: FrameType, event: str, arg: Any) -> Any:
                 if app._loop_thread is None:
                     return super().trace_dispatch(frame, event, arg)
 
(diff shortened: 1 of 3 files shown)


zhangli091011:
Merged the latest upstream/main and resolved the breakpointhook typing conflict by keeping upstream's newer typing fix. The confirm() boolean behavior and regression coverage are preserved; the focused confirm tests (4 cases), Ruff lint/format, and mypy on the touched source files pass.

zhangli091011:
Friendly bump: the branch is clean, all CI checks pass, and the focused confirm() regression tests are green. Could a maintainer review when convenient?

PR #2079, opened 2026-07-25
Avoid redundant renderer resets during application lifecycle

### LINKED ISSUE #1967: `Renderer.reset` is called 3 times during application startup.
This results in some output escape sequences to be produced multiple times without reason, like showing the cursor.

`Renderer.reset()` is called three times when we create/start an application, like ptpython for instance:
- Once `Application.__init__()`-> `Renderer.__init__()` -> `Renderer.reset()`.
- Once `Application.__init__()` -> `Application.reset()` ->`Renderer.reset()`.
- Once `Application._run_async()` -> `Application.__init__()` -> `Application.reset()` ->`Renderer.reset()`.

Each time, we call `self.output.show_cursor()`.

Either we should keep track in the Renderer, which terminal state was changed, and only reset that (cursor shapes changed, cursor hidden/shown, alternate screen entered/left, ...) or we should create the `Renderer` instance only at the point that the application actually starts in `Application._run_async()`, so that we only reset once.

deepakganesh78:
Fixes #1967.

## Summary

- track whether the renderer currently owns terminal state that needs cleanup
- skip renderer resets during `Application` initialization and first startup, when nothing has rendered yet
- avoid the duplicate fallback reset after a successful final render
- add lifecycle regression coverage for startup/shutdown reset counts and a full-screen resize redraw failure that must still quit the alternate screen

## Validation

- `uv run pytest -q tests/test_application.py tests/test_vt100_output.py` — 4 passed
- `uv run prek run --files src/prompt_toolkit/application/application.py src/prompt_toolkit/renderer.py tests/test_application.py` — passed
- `uv run pytest -q tests/` — 148 passed, 5 skipped; 4 Windows/non-console failures reproduced unchanged on `main` (two history-input timing assertions and two `NoConsoleScreenBufferError` failures)
- targeted mypy reports the same existing trace-callback signature errors at `application.py:1119`, `:1123`, and `:1126`
- GitHub Actions unit tests — 158 passed on every Python 3.10–3.14t job; jobs currently stop at the same three pre-existing mypy trace-callback errors reproduced on unmodified `main`

### DIFF
--- a/src/prompt_toolkit/application/application.py
+++ b/src/prompt_toolkit/application/application.py
@@ -416,7 +416,8 @@ class Application(Generic[_AppResult]):
 
         self._background_tasks: set[Task[None]] = set()
 
-        self.renderer.reset()
+        if self.renderer.has_rendered:
+            self.renderer.reset()
         self.key_processor.reset()
         self.layout.reset()
         self.vi_state.reset()
@@ -750,7 +751,8 @@ class Application(Generic[_AppResult]):
                         # _redraw has a good chance to fail if it calls widgets
                         # with bad code. Make sure to reset the renderer
                         # anyway.
-                        self.renderer.reset()
+                        if self.renderer.has_rendered:
+                            self.renderer.reset()
 
                         # Unset `is_running`, this ensures that possibly
                         # scheduled draws won't paint during the following
--- a/src/prompt_toolkit/renderer.py
+++ b/src/prompt_toolkit/renderer.py
@@ -429,7 +429,7 @@ class Renderer:
 
         # Flush output. `disable_mouse_support` needs to write to stdout.
         self.output.flush()
-        self._has_rendered = False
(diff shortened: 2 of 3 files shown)


deepakganesh78:
Addressed the resize/alternate-screen lifecycle review in e2c21a7f.

`Renderer.reset(leave_alternate_screen=False)` now keeps cleanup pending while `_in_alternate_screen` remains active. The new full-screen regression test schedules a resize, forces the resize redraw to fail before `Renderer.render()`, and verifies final cleanup still quits the alternate screen.

Validation:
- `uv run pytest -q tests/test_application.py tests/test_vt100_output.py` — 4 passed
- `uv run prek run --files src/prompt_toolkit/renderer.py tests/test_application.py` — passed
- the regression test fails against the previous behavior with `alternate_screen_quit == 0`

PR #2089, opened 2026-09-02
fix: use character as insert_text for character-valued ANSI_SEQUENCES entries

### LINKED ISSUE #2086: Character-valued `ANSI_SEQUENCES` entries insert their escape sequence instead of the character
`ANSI_SEQUENCES` maps an escape sequence to a key, and a key may be a plain single character rather than a `Keys` member. Such an entry resolves correctly but types its own escape sequence into the buffer.

**Repro** (3.0.53, no other dependencies):

```python
from prompt_toolkit.input.ansi_escape_sequences import ANSI_SEQUENCES
from prompt_toolkit.input.vt100_parser import Vt100Parser

SEQ = "\x1b[27;2;78~"          # what a terminal in xterm modifyOtherKeys=2 sends for Shift+N
ANSI_SEQUENCES[SEQ] = "N"

presses = []
p = Vt100Parser(presses.append)
p.feed(SEQ); p.flush()
print(presses[0].key, repr(presses[0].data))
# -> N '\x1b[27;2;78~'
```

**End to end**, through a real `PromptSession`:

```python
import asyncio
from prompt_toolkit.input.ansi_escape_sequences import ANSI_SEQUENCES
from prompt_toolkit.input import create_pipe_input
from prompt_toolkit.output import DummyOutput
from prompt_toolkit.shortcuts import PromptSession

SEQ = "\x1b[27;2;78~"
ANSI_SEQUENCES[SEQ] = "N"

async def main():
    with create_pipe_input() as pipe:
        pipe.send_text("ab" + SEQ + "cd\r")
        return await PromptSession(input=pipe, output=DummyOutput()).prompt_async()

print(repr(asyncio.run(main())))
# expected: 'abNcd'
# actual:   'ab\x1b[27;2;78~cd'
```

**Mechanism**

- `Vt100Parser._call_handler(key, insert_text)` is called with `insert_text` = the matched bytes, and builds `KeyPress(key, insert_text)`.
- `key_binding/bindings/basic.py:168` binds `Keys.Any` → `self-insert`.
- `key_binding/bindings/named_commands.py:273` — `self-insert` does `insert_text(event.data * event.arg)`.

So insertion is driven by `data`, and `data` is the raw sequence. This is invisible for every entry prompt_toolkit ships, because those resolve to `Keys` members whose bindings never read `data`. It only bites character-valued entries, which the table accepts without complaint.

**Why anyone registers those**

There's no built-in support for xterm `modifyOtherKeys` / the kitty keyboard protocol (#2006 has been open for this since 2025-08-19). The natural workaround is to push the terminal into `CSI >4;2m` and register the resulting sequences as character-valued aliases — at which point every shifted letter types its own escape sequence.

Downstream this is currently handled by monkeypatching `_call_handler`. Hermes Agent has **four independent open implementations of that same patch** — NousResearch/hermes-agent#87785, #88097, #94385, #96133 — plus roughly ten more PRs and six issues against variants of the symptom (Shift+letter, Shift+Space, Shift+symbol, numeric keypad, Cyrillic, Cmd/Super). That seemed like enough duplicated effort to be worth reporting at the source rather than patching around again.

**Ask:** either narrow `data` to the key for character-valued entries, or document that `ANSI_SEQUENCES` values must be `Keys` members. Silently accepting a character and then typing the bytes is the part that costs people time — either resolution fixes that.

Happy to send a PR for whichever shape you'd prefer. Deliberately not opening one first: `data` is read at ~46 sites in the library, including mouse handling that slices it positionally, so which way this should be resolved seems like your call rather than something to presume in a diff.

**Versions:** reproduced on 3.0.52 and 3.0.53, Python 3.11.16, Linux.

---

*Repros written and run by Nyx, a Claude Code agent, on the reporting machine; both are copy-pasteable and were executed as shown.*

dajiaohuang:
## Description

When a character-valued entry is added to `ANSI_SEQUENCES` (e.g., for xterm modifyOtherKeys support), the `KeyPress` data should be the character, not the raw escape sequence. Otherwise, self-insert bindings will insert the escape sequence instead of the character.

## Fixes

Fixes #2086

## Changes

- `src/prompt_toolkit/input/vt100_parser.py`: In `_call_handler`, when `key` is a plain character (not a `Keys` member), use the character as `insert_text` instead of the raw escape sequence
- `tests/test_inputstream.py`: Add regression test `test_character_valued_ansi_sequence`

## Behavior Change

Before: `ANSI_SEQUENCES["\x1b[27;2;78~"] = "N"` would cause `KeyPress("N", "\x1b[27;2;78~")` to be created, and self-insert would insert the raw escape sequence
After: `KeyPress("N", "N")` is created, and self-insert correctly inserts "N"


despotak:
Thanks for picking this up — I filed #2086 and had not expected a fix this quickly.

I have verified the patch rather than just read it, because the issue was terse and I owed
it more evidence than I originally gave.

**The bug reproduces on a clean install**, no downstream patches, prompt_toolkit 3.0.52:

```python
from prompt_toolkit.input.vt100_parser import Vt100Parser
from prompt_toolkit.input.ansi_escape_sequences import ANSI_SEQUENCES

ANSI_SEQUENCES["\x1b[27;2;78~"] = "N"        # Shift+N via xterm modifyOtherKeys=2
out = []; Vt100Parser(out.append).feed("\x1b[27;2;78~")
# KeyPress(key='N', data='\x1b[27;2;78~')
```

`self-insert` inserts `event.data`, so the user gets `^[[27;2;78~` in the buffer instead of
`N`. It is still present on `main` at `583b341` (the 3.0.53 release commit).

**With this PR applied to that commit:**

```
KeyPress(key='N', data='N')
```

**Test suite, run against the cloned source rather than an installed wheel** (`PYTHONPATH=src`,
which matters — pointing pytest at the checkout without it silently tests the installed
copy):

```
main @ 583b341        156 passed
+ this PR            157 passed        the new regression test, no other change
```

And the new test is genuinely red on base — stashing only
`src/prompt_toolkit/input/vt100_parser.py` and keeping the test file fails it, so it pins
the behaviour rather than the implementation.

**One data point on the shape of the fix, since `if not isinstance(key, Keys): insert_text = key`
is broader than the issue strictly requires.** A downstream project has been shipping an
equivalent patch in production since 2026-08-17 with two extra conditions — `len(key) == 1`
and `insert_text.startswith("\x1b")`. I expected that narrowing to matter and it does not:
running both against the same inputs on stock 3.0.52 they agree everywhere I could find to
look —

| input | both produce |
|---|---|
| character-valued entry `ESC[27;2;78~` | `('N', 'N')` |
| plain typed `a` / `7` | `('a', 'a')` / `('7', '7')` |
| a `Keys` entry `ESC[A` | `(Keys.Up, '\x1b[A')` |
| an Alt chord (tuple) `ESC b` | `(Keys.Escape, '\x1b'), ('b', '')` |
| bracketed paste | `(Keys.BracketedPaste, 'hi')` |
| control char `\x03` | `(Keys.ControlC, '\x03')` |

The tuple and bracketed-paste branches return before this line, and for a plain typed
character `key` and `insert_text` are already equal, so the assignment is a no-op there.
I could not construct a case that distinguishes the two. Reporting that as a
non-objection, not a suggestion.

**Why it matters beyond the one-line repro:** every project adding xterm modifyOtherKeys or
kitty CSI-u entries to `ANSI_SEQUENCES` hits this on the first Shift+symbol, because those
protocols report keys whose natural mapping is a plain character rather than a `Keys`
member. That is what the downstream patch above exists for, and it is why the entry point
is worth fixing in core rather than in each consumer.

